PR #14, opened 2020-04-21
test(DIST-142): Embed functional test coverage

antoniofull:
Improves functional test on mobile.
- [x] Provide a **test plan** demonstrating that the code is solid.
- [x] Match the **code formatting** of the rest of the codebase.
- [x] Two other colleagues must **approve** this change.
- [ ] **QA** have tested the solution provided.
- [ ] **Design and UX** must validate this change.

### DIFF
--- a/e2e/cypress-utils.js
+++ b/e2e/cypress-utils.js
@@ -3,6 +3,9 @@
  * @param {Sting} iframe CSS selector
  */
 
+export const WAIT = 1500
+export const LONG_WAIT = 3000
+
 const getIframeDocument = (iframe) => {
   return cy
     .get(iframe)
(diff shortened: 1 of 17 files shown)


e2e/cypress-utils.js
```diff
@@ -86,3 +83,26 @@ export const openAsMobile = (url) => {
     }
   })
 }
+
+export const setupEmbedOnMobile = () => {
+  if (Cypress.isBrowser('chrome')) { // inside of an iframe is testable only in Chrome
+    getIframeBody(IFRAME).find('[data-qa="start-button"]').should('have.text', 'Start').click()
+
+    cy.wait(1500).then(() => {
```
lfons, line 91: Just a small comment. Maybe we could add these waiting times in a const variable so it is easy to adjust them at once if needed.

antoniofull: makes sense, in some cases might not be needed but the main issue is the loading time of the iframe. Let me me have a look later and I'll move them in some variables

antoniofull: fixed now thanks!

antoniofull:
> Looking good. I would like to ask you one thing - can you use the same form, the one without welcome page (https://admin.typeform.com/to/Z9k3bK) for each test? That way it will be more obvious on mobile vs desktop. Thanks :)
on it!

PR #20, opened 2020-04-27
chore(DIST-144): Update Travis npm release config

mathio:
Docs for semantic release say it should be configurable via `release` prop in package.json but it does not work. Passing the argument to the command directly seems to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,9 +83,9 @@ jobs:
         - yarn test:visual:firefox
         - kill $EMBED_PID
     - stage: "Release"
-      name: "Deploy"
+      name: "Deploy Preview to AWS"
       # Deploy to preview preview URL for testing purposes. Only deploy pull requests to 'release' branch.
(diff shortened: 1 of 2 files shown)


prilutskiy:
How about introducing a [`.releaserc`](https://semantic-release.gitbook.io/semantic-release/usage/configuration#configuration-file) file for keeping the semantic-release configuration in it? It would result in a cleaner `package.json`

mathio:
@prilutskiy I can try that. However for some reason the config in `package.json` did not work 🤷‍♂️

mathio:
For some reason the `semantic-release` is not reading the config from a file (neither package.json nor .releaserc), it needs to be passed directly via command line option.

prilutskiy:
@mathio it's a pity to hear that. Seems like they have a bug, because I'm pretty sure I used to get such setup working :)
Nevertheless - approved.

PR #22, opened 2020-04-28
feat(DIST-41): Add event data to onSubmit callback

LexSwed:
[DIST-41](https://jira.typeform.tf/browse/DIST-41)

- adds `takeSubmitEventData` which can be used to extend the details taken from the event. Currently it's just `response_id` passed from stakhanov (PR here: https://github.com/Typeform/stakhanov/pull/3481)
- adds tests for `takeSubmitData`
- makes `onSubmit` handlers to pass event to `takeSubmitEventData` to pass it to provided callback
- adds tests that `event` is passed to the util function `takeSubmitEventData`

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,27 @@ typeformEmbed.makePopup(url, options)
   )
   ```
 
+### `onSubmit` event
+
+Callback function `onSubmit` receives `event` object with next properties:
+
+| Parameter     | Type   | Description        |
+| ------------- | ------ | ------------------ |
+| `response_id` | string | ID of the response |
+
+#### Example:
+
+```javascript
+const reference = typeformEmbed.makePopup(
+  'https://admin.typeform.com/to/PlBzgL',
+  {
+    onSubmit: function (event) {
+      console.log(event.response_id)
+    }
+  }
+)
+```
+
 ## Troubleshooting
 
 ### An element in my page is over the typeform
--- a/src/core/make-fullscreen.js
+++ b/src/core/make-fullscreen.js
@@ -5,7 +5,8 @@ import {
   applyIOSIframeResizeHack,
   replaceExistingKeys,
   redirectToUrl,
-  noop
+  noop,
+  getSubmitEventData
 } from './utils'
 import onMessage from './utils/message-propagation'
 
@@ -28,8 +29,8 @@ export default function makeFullScreen (iframe, url, options) {
 
   iframe.src = appendParamsToUrl(url, replaceExistingKeys(options, queryStringKeys))
 
-  const onFormSubmit = () => {
-    options.onSubmit()
+  const onFormSubmit = (event) => {
+    options.onSubmit(getSubmitEventData(event))
   }
 
   ensureMetaViewport()
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -174,3 +174,7 @@ export const redirectToUrl = event => {
     document.body.removeChild(anchor)
   } catch (e) {}
 }
+
+export const getSubmitEventData = (event) => {
+  return { response_id: event && event.detail ? event.detail.response_id : undefined }
+}
--- a/src/core/utils/index.spec.js
+++ b/src/core/utils/index.spec.js
@@ -144,4 +144,15 @@ describe('Utilities', () => {
       expect(onMessage).not.toHaveBeenCalled()
     })
   })
+
+  describe('getSubmitEventData', () => {
+    test('takes submission info from the event', () => {
+      const detail = { response_id: 'response_id' }
+      expect(utils.getSubmitEventData({ detail })).toEqual({ response_id: detail.response_id })
+    })
+
+    test('returns undefined for empty object', () => {
+      expect(utils.getSubmitEventData({ detail: {} })).toEqual({ response_id: undefined })
+    })
+  })
 })
--- a/src/core/views/mobile-modal.spec.js
+++ b/src/core/views/mobile-modal.spec.js
@@ -4,6 +4,8 @@ import Enzyme, { shallow, mount } from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
 import CustomEvent from 'custom-event'
 
+import * as utils from '../utils'
+
 import MobileModal from './mobile-modal'
 import Iframe from './components/iframe'
 import CloseIcon from './components/close-icon'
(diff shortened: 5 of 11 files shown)


.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   },
   rules: {
     "comma-dangle": ["error", "never"],
-    complexity: 0
+    complexity: 0,
+    "react/jsx-handler-names": 0
```
antoniofull, line 14: 👍

src/core/utils/index.js
```diff
@@ -174,3 +174,9 @@ export const redirectToUrl = event => {
     document.body.removeChild(anchor)
   } catch (e) {}
 }
+
+export const takeSubmitEventData = (event) => {
+  const { landingId } = event.detail || {}
+
+  return landingId && { response_id: landingId }
```
mathio: Why not make the `response_id` camel case (`responseId`)?

src/core/views/mobile-modal.spec.js
```diff
@@ -67,6 +69,17 @@ describe('MobileModal', () => {
     expect(onSubmitSpy).not.toHaveBeenCalled()
   })
 
+  it('passes event data to onSubmit callback', () => {
+    const onSubmitSpy = jest.fn()
+    const takeSubmitEventDataSpy = jest.spyOn(utils, 'takeSubmitEventData')
+    const event = new CustomEvent('form-submit', { detail: { embedId: EMBED_ID } })
+
+    mount(<MobileModal embedId={EMBED_ID} onSubmit={onSubmitSpy} url={URL} />)
```
antoniofull, line 77: wouldn't shallow rendering be faster than mount?

src/core/utils/index.spec.js
```diff
@@ -151,8 +151,7 @@ describe('Utilities', () => {
 
       expect(utils.takeSubmitEventData({ detail })).toEqual({ response_id: detail.landingId })
 
-      expect(utils.takeSubmitEventData({ detail: {} })).toEqual(undefined)
-      expect(utils.takeSubmitEventData({})).toEqual(undefined)
+      expect(utils.takeSubmitEventData({ detail: {} })).toEqual({ response_id: undefined })
```
antoniofull, line 154: I'd use camelCase for these kind of variables

src/core/utils/index.js
LexSwed: I was thinking to match [API](https://developer.typeform.com/responses/reference/retrieve-responses/) schema

mathio:
Please update your commit message based on [guidelines](https://www.notion.so/typeform/Git-Guidelines-ab8a59603f104b928b515cd2f3355455#5c774fdce9da47468dc1ceb10388bc97)

tl;dr `If applied, this commit will <your title here>`

And squash your commits into 1.

src/core/utils/index.js
LexSwed: so that if you make api request, you get `response.response_id` which is equal to `event.response_id` - less cognitive overhead

mathio: Makes sense 👍

src/core/utils/index.js
```diff
@@ -174,3 +174,9 @@ export const redirectToUrl = event => {
     document.body.removeChild(anchor)
   } catch (e) {}
 }
+
+export const takeSubmitEventData = (event) => {
+  const { landingId } = event.detail || {}
+
+  return { response_id: landingId }
```
mathio, line 181: Please comment here once you find out if [landingId is the same as responseId](https://typeform.slack.com/archives/CSZC3BMMH/p1588002748098400)

.eslintrc.js
LexSwed: okay, fully agree, dumb way of fighting the system

mathio: I know we have a lint warning because of this. You can try to fix that if you want (I tried in past but had issues with this one) but its fine to leave it as is since its out of scope.

antoniofull: You could also disable it just for the file that causes the warning

README.md
```diff
@@ -117,6 +117,27 @@ typeformEmbed.makePopup(url, options)
   )
   ```
 
+### `onSubmit` event
+
+Callback function `onSubmit` receives `event` object with next properties:
+
+| Parameter       | Type   | Description          |
+| --------------- | ------ | -------------------- |
+| `submission_id` | string | ID of the submission |
```
mathio, line 126: This does not seem to match with the code. Its named `response_id` there.

LexSwed: good catch 👍

mathio:
I have a proposal. 

1) We merge this. But before we do, let's call the returned ID a `landingId` since thats what this is. We can deal with other implications later. We could ask results team extend API to allow retrieving responses by `landingId`, but thats for later.

2) We dont merge this since `landingId` is kind of useless.

3) We dont merge this and create a spike ticket to actually return `responseId` here.

@LexSwed @Gotusso WDYT?

LexSwed:
> I have a proposal.
> 
> 1. We merge this. But before we do, let's call the returned ID a `landingId` since thats what this is. We can deal with other implications later. We could ask results team extend API to allow retrieving responses by `landingId`, but thats for later.
> 2. We dont merge this since `landingId` is kind of useless.
> 3. We dont merge this and create a spike ticket to actually return `responseId` here.
> 
> @LexSwed @Gotusso WDYT?

I returned to my team, so would ask someone to take over the task.

I'm into 2, I don't know how many customers we're satisfying with it.
I'm into 1st, `landingId` is documented in response structures, so it's not something new for devs, so we could have an endpoint. and then we could deprecate it. However, if responses are added to DB asynchronously, we can have a case when someone requests response by `landingId` right after form submit, but submission is still in the queue and hence doesn't exists in DB. so need to ping BE devs: https://typeform.slack.com/archives/CSZC3BMMH/p1588065761099600

src/core/utils/index.js
LexSwed: we need centralised place for this :D 
but @pchiwan said they wanted to introduce `response_id` in `stakhanov`, it's in their backlog. so the issue will be solved as soon as they have it

mathio: If this is in their backlog (and it is possible to retrieve `response_id` in stakhanov) I would wait for that.

prilutskiy:
Embed script preview available: https://embed.typeform.com/f6809c62d25be9af3ebb2eaf556700dcea59fbd0/embed.js

prilutskiy:
Embed script preview available: https://embed.typeform.com/de321c2a51ffa30af4c68ffa1511ddb39de5ac62/embed.js

mathio:
I updated this PR and opened a [related PR in stakhanov](https://github.com/Typeform/stakhanov/pull/3481). Now actual `response_id` is passed, not `landing_id`.

@LexSwed and @Typeform/distribution please re-review.

src/core/make-fullscreen.js
```diff
@@ -5,7 +5,8 @@ import {
   applyIOSIframeResizeHack,
   replaceExistingKeys,
   redirectToUrl,
-  noop
+  noop,
+  takeSubmitEventData
```
antoniofull, line 9: I would have called `getSubmitEventData` 😄

prilutskiy:
Embed script preview available: https://embed.typeform.com/0643980aa29da76e2dcd4ce821916f49b3d895c0/embed.js

travis-tf:
Embed script preview available: https://embed.typeform.com/a389e6f5a0ea642182d4e5272870d9914fb5ccbc/embed.js

travis-tf:
Embed script preview available: https://embed.typeform.com/cf81b1083ef5d80cf9efd41f04080b720335dff7/embed.js

PR #28, opened 2020-05-06
feat: Added onReady & onClose to the API

prilutskiy:


mathio:
Could you please add tests for this? It would be great to have this tested in functional tests.

prilutskiy:
@mathio yep, for sure 👍

package.json
```diff
@@ -26,24 +26,24 @@
...
+    "test:visual:chrome": "codeceptjs --config codecept-visual.conf.js run --steps --grep @desktop",
+    "test:visual:firefox": "codeceptjs --config codecept-visual.conf.js run --steps --profile firefox --grep @desktop",
+    "test:visual:mobile": "codeceptjs --config codecept-visual.conf.js run --steps --grep @mobile",
+    "test:functional:debug": "cypress open",
     "test:functional": "yarn test:functional:chrome && yarn test:functional:firefox",
-    "test:functional:firefox": "yarn cypress run --browser firefox",
-    "test:functional:chrome": "yarn cypress run --browser chrome",
-    "prepublish": "yarn run lib",
+    "test:functional:firefox": "cypress run --headless --browser firefox",
+    "test:functional:chrome": "cypress run --headless --browser chrome",
```
mathio: If you run `yarn cypress` it should be always taken from the `node_modules`. If you run just `cypress` and you happen to have cypress installed globally which one is used?

The reason I like to have `yarn run` is that it explicitly tells you are running a custom script from `package.json`. For example if you had a script named `version` you would not be able to run it using `yarn version` since there is a built-in yarn command like that.

prilutskiy: > which one is used

the local one is used, and if it's not in node_modules - then the globally installed one is used.

I personally don't like `run` for its explicitness; I'm a big fan of doing `yarn` instead of `yarn install`, `npm i` instead of `npm install`, etc. However, since this topic really smells like a holywar, I'll just leave these lines as I've found them :)

mathio: > the local one is used
👍 

> smells like a holywar
I am fine with new ideas. I too write `yarn install` in terminal because its short. However the `package.json` is written just once 😃 I am fine with changing this if the rest of the team shares your view on this.
WDYT @Gotusso @lfons @antoniofull ?

.travis.yml
```diff
@@ -10,6 +10,7 @@ python:
   - "3.6"
 
 addons:
+# https://stackoverflow.com/questions/57903415/travis-ci-chrome-62-instead-of-77
```
prilutskiy, line 13: 🤦

package.json
```diff
@@ -72,6 +72,7 @@
     "jsdom": "^11.6.2",
     "react": "16.12.0",
     "react-dom": "16.12.0",
+    "regenerator-runtime": "^0.13.5",
```
prilutskiy, line 75: needed for async tests because babel is not up to date.

package.json
Gotusso: If you ask me we should favour readability over shortness because we will read it a thousand times and yada yada yada 🤷‍♂️ But yeah, holy war, and seems a bit out of topic for this PR.

README.md
```diff
@@ -140,15 +154,15 @@ Although we have no hard limit, we recommend having a height of at least 350px.
 We use `position: fixed` to position our modal relative to its containing block established by the viewport. If one of the modal ancestors has a `transform`, `perspective`, or `filter` css property set to something other than `none` the positioning will be relative to it and probably not visible by the user.
 
 ## Tests
-In order to run visual tests, it is need an applitools key. 
-- Add a `.env` file in your root, you can look at the `.env.example` 
+In order to run visual tests, it is need an applitools key.
```
mathio, line 157:
```suggestion
In order to run visual tests, you need an applitools key.
```

package.json
```diff
@@ -21,29 +21,29 @@
...
+    "test:unit:watch": "yarn jest --watch",
+    "test:unit:coverage": "yarn jest --coverage",
+    "test:visual": "yarn run test:visual:install && yarn run test:visual:chrome && yarn run test:visual:firefox && yarn run test:visual:mobile",
     "test:visual:install": "yarn selenium-standalone install --silent",
     "test:visual:chrome": "yarn codeceptjs --config codecept-visual.conf.js run --steps --grep @desktop",
     "test:visual:firefox": "yarn codeceptjs --config codecept-visual.conf.js run --steps --profile firefox --grep @desktop",
     "test:visual:mobile": "yarn codeceptjs --config codecept-visual.conf.js run --steps --grep @mobile",
     "test:functional:debug": "yarn cypress open",
-    "test:functional": "yarn test:functional:chrome && yarn test:functional:firefox",
+    "test:functional": "yarn run test:functional:chrome && yarn run test:functional:firefox",
```
mathio, line 34: Good catch 👍

PR #34, opened 2020-05-19
fix: Use shared travis config

LexSwed:
Try to use shared config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
+version: ~> 1.0
+import: Typeform/travis-shared-config:package-config.yml@master
 os: linux
 dist: xenial
 
 language: node_js
-
 node_js:
   - "12"
 


mathio:
Please include a ticket number in commit message, we need more context for this. This seems to be just a test, do you need to get this merged?

Are we able to have single travis for all projects since each is configured differently? From what I've seen embed has different Travis jobs than other projects since its open source.

LexSwed:
> Please include a ticket number in commit message, we need more context for this. This seems to be just a test, do you need to get this merged?
> 
> Are we able to have single travis for all projects since each is configured differently? From what I've seen embed has different Travis jobs than other projects since its open source.

Hey @mathio !

We have this [big epic](https://jira.typeform.tf/browse/FEAR-25). Basically the idea is to migrate all `npm` packages from `npm` registry to `GitHub` registry, so that nobody needs to go and create own personal `npm` tokens which we cannot dispose even if a person already left Typeform, because we don't know which `npm` token belongs to whom. So we need to migrate all existing packages to GitHub registry + publish to both registries, until the migration is finished and we can just remove all tokens from npm. Then everyone should setup their local env and repos to fetch `@typeform/` packages from `GitHub` registry. However, with Open Source libraries it's different because we need them to be published to `npm`, but be accessible for us (those who use `GitHub` registry for `@typeform` packages). So we need to publish to both registries. Initially we wanted to go with a common config for public packages (to avoid any confusions for future OS libraries, but it seems too complicated with `travis`, so we're going to update travis configs to publish to both registries manually.

LexSwed:
Closing in favour of https://github.com/Typeform/embed/pull/35

PR #35, opened 2020-05-20
fix: Publish to both registries

LexSwed:
We have this [big epic](https://jira.typeform.tf/browse/FEAR-25). Basically the idea is to migrate all `npm` packages from `npm` registry to `GitHub` registry, so that nobody needs to go and create own personal `npm` tokens which we cannot dispose even if a person already left Typeform, because we don't know which `npm` token belongs to whom. So we need to migrate all existing packages to GitHub registry + publish to both registries, until the migration is finished and we can just remove all tokens from npm. Then everyone should setup their local env and repos to fetch `@typeform/` packages from `GitHub` registry. However, with Open Source libraries it's different because we need them to be published to `npm`, but be accessible for us (those who use `GitHub` registry for `@typeform` packages). So we need to publish to both registries. Initially we wanted to go with a common config for public packages (to avoid any confusions for future OS libraries, but it seems too complicated with `travis`, so we're going to update `travis` configs to publish to both registries manually.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -7,6 +7,12 @@
     "@semantic-release/release-notes-generator",
     "@semantic-release/changelog",
     "@semantic-release/npm",
+    [
+      "@semantic-release/exec",
+      {
+        "successCmd": "yarn run publish:github"
+      }
+    ],
     "@semantic-release/git",
     "@semantic-release/github"
   ]
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "copy:demo": "yarn copyfiles -f demo/* dist",
     "build": "yarn run dist && yarn run lib && yarn run copy",
     "dist": "yarn run clean:dist && yarn webpack --config webpack.config.dist.js",
-    "lib": "yarn run clean:lib && yarn webpack --config webpack.config.lib.js"
+    "lib": "yarn run clean:lib && yarn webpack --config webpack.config.lib.js",
+    "publish:github": "npm config set '//npm.pkg.github.com/:_authToken' $GH_TOKEN && npm publish --registry https://npm.pkg.github.com"
   },
   "devDependencies": {
     "@applitools/eyes-webdriverio": "^5.7.2",
@@ -50,6 +51,7 @@
     "@commitlint/config-conventional": "^8.3.4",
     "@semantic-release/changelog": "^5.0.1",
     "@semantic-release/commit-analyzer": "^8.0.1",
+    "@semantic-release/exec": "^5.0.0",
     "@semantic-release/git": "^9.0.0",
     "@semantic-release/github": "^7.0.6",
     "@semantic-release/npm": "^7.0.5",
--- a/yarn.lock
+++ b/yarn.lock
@@ -951,6 +951,18 @@
   resolved "https://registry.yarnpkg.com/@semantic-release/error/-/error-2.2.0.tgz#ee9d5a09c9969eade1ec864776aeda5c5cddbbf0"
   integrity sha512-9Tj/qn+y2j+sjCI3Jd+qseGtHjOAeg7dU2/lVcqIQ9TV3QDaDXDYXcoOHU+7o2Hwh8L8ymL4gfuO7KxDs3q2zg==
 
+"@semantic-release/exec@^5.0.0":
+  version "5.0.0"
+  resolved "https://registry.yarnpkg.com/@semantic-release/exec/-/exec-5.0.0.tgz#69c253107a755dabf7c262d417269d099f714356"
+  integrity sha512-t7LWXIvDJQbuGCy2WmMG51WyaGSLTvZBv9INvcI4S0kn+QjnnVVUMhcioIqhb0r3yqqarMzHVcABFug0q0OXjw==
+  dependencies:
+    "@semantic-release/error" "^2.1.0"
+    aggregate-error "^3.0.0"
+    debug "^4.0.0"
+    execa "^4.0.0"
+    lodash "^4.17.4"
+    parse-json "^5.0.0"
+
 "@semantic-release/git@^9.0.0":
   version "9.0.0"
   resolved "https://registry.yarnpkg.com/@semantic-release/git/-/git-9.0.0.tgz#304c4883c87d095b1faaae93300f1f1e0466e9a5"


.travis.yml
```diff
@@ -35,7 +34,6 @@ env:
 before_install:
   - export AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY
   - export AWS_SECRET_ACCESS_KEY=$AWS_SECRET_KEY
-  - npm config set '//registry.npmjs.org/:_authToken' ${NPM_TOKEN}
```
LexSwed, line 38: you don't need it since you don't use private `@typeform` packages (and shouldn't as it's OS)

mathio: Very soon we will want to use Jarvis for deployments. Not as dependency, just installing it locally inside Travis job.
https://jira.typeform.tf/browse/DIST-242

mathio:
Maybe our eslint-config is not published correctly as public? I think you get 404 for private packages without npm token.
```
error An unexpected error occurred: "https://registry.yarnpkg.com/@typeform/eslint-config/-/eslint-config-1.12.0.tgz: Request failed \"404 Not Found\"".
```

LexSwed:
> Maybe our eslint-config is not published correctly as public? I think you get 404 for private packages without npm token.
> 
> ```
> error An unexpected error occurred: "https://registry.yarnpkg.com/@typeform/eslint-config/-/eslint-config-1.12.0.tgz: Request failed \"404 Not Found\"".
> ```

yeah, working on it. Weirdly the package had `"publishConfig": { "access": "restricted" }` in `package.json`

.travis.yml
LexSwed: I see 🤔

prilutskiy:
Embed script preview available: https://embed.typeform.com/c8029d7031012dc5a243f5af04b89561aefcebaa/embed.js

prilutskiy:
Embed script preview available: https://embed.typeform.com/da43c79820cd2291fd570c11014aa9733690981e/embed.js

LexSwed:
@mathio can you take a look again?

.travis.yml
```diff
@@ -96,6 +95,10 @@ jobs:
       if: branch = master AND type = push
       script:
         - yarn semantic-release
+        - npm config set '//npm.pkg.github.com/:_authToken' $GH_TOKEN
+        - npm config set registry https://npm.pkg.github.com/@typeform
+        - npm publish
```
mathio, line 100: Does this publish the same code to both registries? `semantic-release` takes care of versioning and only releases when there are any changes (commit message starts with `fix`, `feat`, etc.). I think `npm publish` might work a bit differently.

LexSwed: Nice catch, didn't think of that, yes, it'll try to publish with the version specified in `package.json`, and since `semantic-release` didn't make any changes there, it'll take current version. While here it's not that much of an issue since `package.json` has actual `version` (so it'll just fail because it won't be able to publish same version), in other packages it will publish whatever is specified in `package.json#version` 🤔

LexSwed: 💩

mathio: We do not update the version in `package.json`, in most projects it has a value of `"version": "0.0.0-semantic-release",` and we let semantic release take care of it.

I see the version is changed here by @prilutskiy but he was resolving some issues with `semantic-release` so this might be just a temporary thing. Am I correct?

LexSwed: okay, yeah, I see that you don't commit `package.json` after release, I only saw changed version in `package.json`. I can still think of another safe solution: use plugin with shortly config:
`["@sematic-release/exec", { "successCmd": "npm publish" }]`

mathio: Previously we were using outdated version of semantic release. Max should be having a look to let us know if your approach works.

prilutskiy: A couple of notes:
1. Version in package.json is now handled automagically, by semantic release
2. To publish the same thing to an additional channel, I recommend using semantic-release also. This is how its config looks like currently:

```
{
  "branches": [
    "master"
  ],
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    "@semantic-release/changelog",
    "@semantic-release/npm",
    "@semantic-release/git",
    "@semantic-release/github"
  ]
}
```

In my opinion, to achieve the goal it makes sense to add something like `@semantic-release/github-package-registry`, in addition to the npm plugin specified here. I'm not sure such plugin exists already, but in case it does not - there's a thing called [@semantic-release/exec](https://github.com/semantic-release/exec)

prilutskiy: UPD:

Found [something](https://github.com/gr2m/semantic-release-github-npm-registry), but maybe we'd just go with our custom approach, if the tool is not that good

prilutskiy:
Embed script preview available: https://embed.typeform.com/c6da96e1de1e82e4168e98cdc55ec7bb38631f81/embed.js

.travis.yml
LexSwed: @prilutskiy I think there is no proper solution that does what we need
I'm using just `npm publish` because I want to rely on `@semantic-release/npm` complete everything for me (e.g. if `npm pack` is needed to generate `tar` with updated `version` and `dist` fields updated in `package.json`, then just run [`npm publish`](https://github.com/semantic-release/npm/blob/680a82774118ecd935e07c2d480e1f5289c07180/lib/publish.js#L23)

prilutskiy:
Embed script preview available: https://embed.typeform.com/491534139e2043cca249917fd4ae31a75991741b/embed.js

prilutskiy:
Embed script preview available: https://embed.typeform.com/7e3401511f91c1867a7a65f8eed63241808764d9/embed.js

prilutskiy:
> However, with Open Source libraries it's different because we need them to be published to npm, but be accessible for us (those who use GitHub registry for @typeform packages). So we need to publish to both registries.

@LexSwed I think there should be a better solution for open source projects. I believe pushing them to both registries is not the best idea: potentially it leads to inconsistency and issues that quite hard to debug. Since the GitHub registry is just a proxy to the original npmjs registry - how about leaving open-source packages just where they are now, in npmjs?

LexSwed:
That's why we do it PR way: we only trigger publish to GitHub registry when `semantic-release` published. The only risk for this repo is the package from `github` registry (used only by typeformers) being different from the one from `npm` registry.
And since we set up the registry for all `@typeform` scoped packages to be `github`, installation will fail if there is no such package in the registry. It won't proxy request to npm since we explicitly say to download `@typeform` packages from `github`.

.travis.yml
```diff
@@ -101,4 +101,4 @@ jobs:
       script:
         - yarn build
         - pip install --user awscli
-        - aws s3 cp ./dist s3://$AWS_ASSETS_BUCKET --recursive --acl public-read
+        - aws s3 cp ./dist s3://$AWS_ASSETS_BUCKET --recursive --acl public-read
```
mathio, line 104: Please dont remove empty line at the end of the file.

mathio:
@LexSwed looking good, just 2 minor things before we can merge this PR:

1. please [squash your commits into 1](https://www.notion.so/typeform/Git-Branches-ff30bc3692204ff69b88c1a8c64c2f26#a44aa5d7f06140aa8c13a747cefc8264)
2. update your commit message to include ticket number (we want to trace tickets via `git blame`) and prefix the commit message `feat`, since this is a new feature rather than a bugfix (I would suggest `chore` since this is build change but I am not sure if semantic release will release a new version for this prefix)

LexSwed:
👍 on it

Maybe we can go with `chore`, without publishing, it should work 😅 only issue can be is if your GH_TOKEN is outdated in Travis, then you'll need to run `tfcli` command to update env variables.

LexSwed:
git is hard 
the fuck is going on with these force pushes

prilutskiy:
Embed script preview available: https://embed.typeform.com/f6a70201d849383f219ee0519f93fcb81b2ef94d/embed.js

mathio:
@LexSwed I see the author of the commit is `semantic-release-bot <semantic-release-bot@martynus.net>` now. Please update the author and I think this can be merged:

```
git commit --amend --author="John Doe <john@doe.org>"
```

prilutskiy:
Embed script preview available: https://embed.typeform.com/fd98d3208651712ad50e87025222be0c9b4b2daa/embed.js

PR #41, opened 2020-06-04
chore(DIST-242): Use Jarvis to deploy

mathio:
Install [Jarvis](https://github.com/Typeform/jarvis) package in Travis deploy job and use it for deployment to AWS. Since [Jarvis](https://github.com/Typeform/jarvis) is a private package we do not add it to dependencies in `package.json` or the installation for people from outside of the company would fail. It is installed only in the Travis CI deploy job and changes are not committed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,8 +82,8 @@ jobs:
       script:
         - yarn build
         - pip install --user awscli
-        - aws s3 cp ./dist s3://$AWS_ASSETS_BUCKET/$TRAVIS_COMMIT --recursive --acl public-read
-        - ./scripts/comment-preview-link
+        - yarn add @typeform/jarvis
+        - DEBUG=jarvis yarn run jarvis deploy --path dist --preview --notify-preview
     - name: "Release to NPM"
       if: branch = master AND type = push
       script:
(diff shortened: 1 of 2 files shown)


travis-tf:
*[BOT]* Preview available, append `?embedVersion=c46c14e888d969c319369c1168750ceae5444de3` to a form url to check it out.

travis-tf:
*[BOT]*

travis-tf:
*[BOT]*

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/c46c14e888d969c319369c1168750ceae5444de3/embed.js

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/e11edc30812b0e94a7ee9725ff15141fb3a7c33b/embed.js

.travis.yml
```diff
@@ -82,8 +82,8 @@ jobs:
       script:
         - yarn build
         - pip install --user awscli
-        - aws s3 cp ./dist s3://$AWS_ASSETS_BUCKET/$TRAVIS_COMMIT --recursive --acl public-read
-        - ./scripts/comment-preview-link
+        - yarn add @typeform/jarvis
+        - DEBUG=jarvis yarn run jarvis deploy --path dist --preview --notify-preview
```
prilutskiy, line 86: just a side note - we also have a lib folder generated along with the dist. We don't need to upload it anywhere, do we?

mathio: I think that is used when you install the lib via npm. We are not deploying it right now and everything works :)

prilutskiy: @mathio cool! Mentioned that just in case :)

mathio: Yes, we should always point out if anything seems odd. More 👀  = higher code quality.

PR #44, opened 2020-06-12
fix(DIST-193): Fix for autoOpen backward compatibility

mathio:
Also add docs for behavioral triggers functionality.


mathio:
@Gotusso good catch. I updated the docs and also marked the "autoOpen" option as deprecated.

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/c1fd0c6d593c9550985d1699b3757a0053e9798d/embed.js

PR #54, opened 2020-07-09
fix(DIST-306): Add side panel popup type

mathio:
For popup mode "side_panel" you can also set width and height of the popup. It needs to specify "container" in order for slide animation to work correctly.

https://jira.typeform.tf/browse/DIST-306


travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/9162a618ffe0d2dd760c7ee8b7cbefba79a14e09/embed.js

PR #62, opened 2020-07-31
fix(DIST-366): Fix icon behavior when using custom launch options

lfons:
- [x]  Link to the issue - if any [issue number here]()

### DIFF
--- a/src/core/make-popup.js
+++ b/src/core/make-popup.js
@@ -134,7 +134,7 @@ export default function makePopup (url, userOptions) {
     open (event) {
       const { currentTarget } = event || {}
       const currentUrl = currentTarget && currentTarget.href ? currentTarget.href : url
-      const icon = currentTarget && currentTarget.querySelector('span.icon')
+      const icon = (currentTarget && currentTarget.querySelector('span.icon')) || document.querySelector('a.typeform-share.button > span.icon')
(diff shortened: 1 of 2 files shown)


travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/1a676f3aff8609356a2eb8a5393820ab155f27ca/embed.js

lfons:
> How about classes instead of inline styles? Looks more readable, imo.

In this specific scenario, I just copied the code generated by the Typeform Share page to closely mimic a real use-case. But yes, in general, we should be using classes :)

src/core/make-popup.js
```diff
@@ -134,7 +134,7 @@ export default function makePopup (url, userOptions) {
     open (event) {
       const { currentTarget } = event || {}
       const currentUrl = currentTarget && currentTarget.href ? currentTarget.href : url
-      const icon = currentTarget && currentTarget.querySelector('span.icon')
+      const icon = (currentTarget && currentTarget.querySelector('span.icon')) || document.querySelector('a.typeform-share.button > span.icon')
```
mathio, line 137: What if there are more buttons on the page? I think we need to use `currentTarget` somehow. Maybe send a custom `event` when opening it automatically?

PR #70, opened 2020-08-13
feat(DIST-390): Added node.js support

### LINKED ISSUE #68: [BUG] Window undefined error in next.js
**Describe the bug**
Trying to use library in next.js and getting window is undefined when using makeWidget within useEffect which is client side rendered only

**To Reproduce**
See basic example in repo:
https://github.com/conorshaw/typeform-next

**Expected behavior**
Shouldn't error out with window undefined when client side rendered

**Current behavior**
Throws window is undefined error

prilutskiy:
resolves #68

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -48,6 +48,9 @@ function getWebpackConfig () {
       // UMD supports both NPM package format and
       // global variable format, when used from a CDN
       libraryTarget: 'umd',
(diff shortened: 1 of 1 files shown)


travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/e97f92c30d25fd5085c59debb6d76b30658ada2d/embed.js

PR #74, opened 2020-09-20
fix(DIST-453): Fix page height calculation

mathio:
There is an issue when host page has CSS.

```
html {
    overflow-y: scroll;
    overflow-x: hidden;
    height: 100%;
}
```

In that case we consider the page to have the same height as viewport, which is kind of correct (the <html> is 100%) but not really correct since page content overflows from <html>. See [DIST-453](https://jira.typeform.tf/browse/DIST-453) for details.

- [x]  Link to the issue - [DIST-453](https://jira.typeform.tf/browse/DIST-453)


travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/5c231e2bbc21f61b49462dcfaf5fefa980b41060/embed.js

prilutskiy (approved):
LGTM 👍

PR #76, opened 2020-09-30
fix(DIST-459): Remove iOS resize hack that is no longer needed

mathio:
Some forms are not displayed on iOS (see [DIST-459](https://jira.typeform.tf/browse/DIST-459)) because the <iframe> has width=0. This is due to an iOS hack introduced in 2017 (see [SB-731](https://jira.typeform.tf/browse/SB-731)) which seems to be no longer relevant.

This issue does not affect all forms, so it might be a conflict with host page CSS.

TODO:
- [x] more testing to confirm the hack can be removed

### DIFF
--- a/src/core/make-fullscreen.js
+++ b/src/core/make-fullscreen.js
@@ -1,8 +1,6 @@
 import {
   appendParamsToUrl,
   ensureMetaViewport,
-  applyIOSFooterHack,
-  applyIOSIframeResizeHack,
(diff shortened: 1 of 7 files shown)


prilutskiy (approved):

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/fabacdf92689841ae57ab082f43e69b1542aefee/embed.js

Issue #24, opened 2020-04-28
Large bundle size


bogdansoare:
This library comes at a whopping 113.1kB minified https://bundlephobia.com/result?p=@typeform/embed@0.16.0 

That is not very good for just showing a form on a website.

bogdansoare:
```
info Direct dependencies
└─ @typeform/embed@0.16.0
info All dependencies
├─ @emotion/stylis@0.7.1
├─ @emotion/unitless@0.6.7
├─ @typeform/embed@0.16.0
├─ classnames@2.2.6
├─ create-emotion-styled@9.2.8
├─ create-emotion@9.2.12
├─ custom-event@1.0.1
├─ immutability-helper@2.9.1
├─ preact-compat@3.19.0
├─ preact-context@1.1.4
├─ preact-render-to-string@3.8.2
├─ preact-transition-group@1.1.1
├─ preact@8.5.3
├─ scrollbar-width@3.1.1
├─ spin.js@2.3.2
├─ standalone-react-addons-pure-render-mixin@0.1.1
├─ stylis-rule-sheet@0.0.10
└─ stylis@3.5.4
```

That's a lot of dependencies.
My advice, drop css-in-js, just use css. Css-in-jss is not well suited for libraries unless both the library and the app consuming the library use the same css-in-js solution.

Second, have 2 versions of the lib, a vanilla one without preact and a react one that's a wrapper on top of the vanilla one, since react it's the most popular library.

Gotusso:
Hey @bogdansoare thanks for reaching out! Long story short, you're absolutely right, the bundle is too big for what the library does today. This also loads another bundle once the Typeform shows up that it's bigger, and it's impacting user experience the most, so we're focusing on that one first. After that, we will also check how we can shave some size here. 

In the meantime, I'm going to leave the issue open to keep collecting feedback, and as a channel for the improvements we do related to this issue.

bogdansoare:
Hey @Gotusso thank you for the response, glad to hear that you're working on it.
I really like typeform and would be happy if both developers and users will have a nice experience using this library.

All the best.

prilutskiy:
As of now, we reduced the bundle size by 27%. In the nearest future, we'll also be optimizing a bunch of other stuff, so the bundle will become even tinier.

jellevandevelde:
I'm struggling with the same issue here. The bundle size as of date is still very large.
https://bundlephobia.com/result?p=@typeform/embed@0.25.1

[...]

PR #88, opened 2020-10-28
perf: Move react, react-dom to peer deps

osdiab:

react and react-dom shouldn't be pinned so i dont need to install multiple versions of react to use this package

tests ran, just make sure it's still loading as you expect.

- [x]  Link to the issue - if any #24

this doesn't close that issue, it's just a small part of it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,8 @@
     "jest": "^26.0.1",
     "jsdom": "^16.2.2",
     "preact": "^10.4.4",
+    "react": "16.13.1",
+    "react-dom": "16.13.1",
     "regenerator-runtime": "^0.13.5",
     "selenium-standalone": "^6.17.0",
     "semantic-release": "^17.0.8",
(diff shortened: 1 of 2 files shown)


prilutskiy (approved):
Great job!

osdiab:
That was dev dependencies, you need it there because the tests depend on react being present.

mathio:
You are correct. But if we keep it as dev dep it is included in the final bundle:
Screenshot 2020-10-28 at 12 37 19

A sample nextjs app would I am testing this in still has 2 copies of react:
Screenshot 2020-10-28 at 12 37 19

Maybe there is something wrong with our build process 🤔

mathio:
I checked again and it looks like I was not installing it correctly. Your PR looks good as is, I removed my commit.

Lets merge it!

mathio:
Sorry for the confusion. The change was released and works as expected. Thank you for the PR!

PR #92, opened 2020-10-30
Pull request bundle size (actions)

prilutskiy:


mathio:
I am missing the info on which bundle size is or which lib (pure / standard). Now it looks weird. Can you include lib name in the comment?

prilutskiy:
@mathio  https://github.com/chronotruck/webpack-stats-diff-action/pull/9 you can support me in the PR ;)

PR #94, opened 2020-11-04
chore: add segment write key

otorrillas:
Adds `SEGMENT_WRITE_KEY`, so it sends deployment events correctly via `jarvis`


otorrillas:
No longer required, added via Travis Env Variables.

PR #99, opened 2020-11-13
feat(DIST-512): Add "size" option for "popup" embed mode

mathio:
Embed lib accepts "size" option:
* as `size` prop in `options` passed to `makePopup` method
* as `data-size` element attribute

Popup size is calculated relative to window size as `${size}% - 80px`. Default value is `100` (full screen).


src/core/make-popup.js
```diff
@@ -31,6 +31,7 @@ import { handleAutoOpen } from './utils/popup-auto-open'
 const DEFAULT_DRAWER_WIDTH = 800
 const DEFAULT_POPUP_WIDTH = 320
 const DEFAULT_POPUP_HEIGHT = 500
+const DEFAULT_POPUP_SIZE = 100
```
lfons, line 34: Maybe it will make sense to indicate that this value is a percent (%). So you can pass the value as '75%' and not as just '75'. However, I see that for the other units HEIGHT & WIDTH, we are using pixels and we are not providing the units.

Maybe we could just add a comment in the code to specify that this 100 (DEFAULT_POPUP_SIZE) is a percentage and NOT pixels like the other ones could do. Or we could discuss as a team if we want to be able to specify the units on a further refactor.

lfons (approved):

src/core/make-popup.js
mathio: Since we are not using units I think we should omit it here as well. Maybe we could rename this constant to `DEFAULT_POPUP_SIZE_PERCENT`?

antoniofull: I think renaming the constant is a good approach, it makes it clear what unit is using

prilutskiy (approved):

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/27f358cb67118bdce87f3728ef53e72109670379/embed.js

PR #102, opened 2020-11-18
chore: Added stale bot config

prilutskiy:


travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/073a34ffc9b59357a35d189f4a0f73bd37a01110/embed.js

mathio:
Since we are not notified when someone posts new issue (at least I am not, even though I have the repo starred and watched) I would wait with this until we have a notifications for new issues in our slack channel. Otherwise we might close an issue before we have a chance to see it.

.github/stale.yml
```diff
@@ -0,0 +1,61 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
```
mathio, line 4: I would go with lower setting, I think 7 days of inactivity should be enough.

lfons:
Is there a way to test it?

prilutskiy:
@lfons only after it's merged

.github/stale.yml
prilutskiy: Done

PR #105, opened 2020-11-23
chore(DIST-549): Add minumum size for popups on desktop

antoniofull:
Adds a minimum size for Embed popups.
[Reference Ticket DIST-549](https://jira.typeform.tf/browse/DIST-549)
I went for the easiest path, and just adding the values inside the desktop popup component
- [x]  Link to the issue - if any [DIST-549](https://jira.typeform.tf/browse/DIST-549)

### DIFF
--- a/src/core/views/popup.js
+++ b/src/core/views/popup.js
@@ -39,6 +39,10 @@ const BaseWrapper = styled.div`
   position: ${p => (p.isContained ? 'absolute' : 'fixed')};
   max-width: 100%;
   z-index: 10001;
+  @media screen and (min-width: 1024px) {
+    min-width: 360px;
+    min-height: 360px;
+  }
 `
 
 const Overlay = styled.div`
@@ -58,23 +62,13 @@ const Overlay = styled.div`
 `
 
 const popupWrapper = styled(BaseWrapper)`
-  ${p => {
-    const offset = (100 - p.size) / 2
-    if (p.isContained) {
-      return `
-        width: calc(${p.size}% - 80px);
-        height: calc(${p.size}% - 80px);
-        top: calc(${offset}% + 40px);
-        left: calc(${offset}% + 40px);
-      `
-    }
-    return `
-      width: calc(${p.size}% - 80px);
-      height: calc(${p.size}% - 80px);
-      top: calc(${offset}% + 40px);
-      left: calc(${offset}% + 40px);
-    `
-  }}
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  ${({ size }) => size && `
+    height: calc(${size}% - 80px); 
+    width: calc(${size}% - 80px); 
+  `}  
   transition: all 300ms ease-out;
 `
 
--- a/src/core/views/popup.js
+++ b/src/core/views/popup.js
@@ -39,10 +39,8 @@ const BaseWrapper = styled.div`
   position: ${p => (p.isContained ? 'absolute' : 'fixed')};
   max-width: 100%;
   z-index: 10001;
-  @media screen and (min-width: 1024px) {
-    min-width: 360px;
-    min-height: 360px;
-  }
+  min-width: 360px;
+  min-height: 360px;
 `
 
 const Overlay = styled.div`


mathio (commented):
It looks like the popup position is no longer centered on small screens:

Screenshot 2020-11-23 at 15 21 36

antoniofull:
> It looks like the popup position is no longer centered on small screens:
> 
> Screenshot 2020-11-23 at 15 21 36

Good catch! Fixing it now thanks

src/core/views/popup.js
```diff
@@ -39,6 +39,10 @@ const BaseWrapper = styled.div`
   position: ${p => (p.isContained ? 'absolute' : 'fixed')};
   max-width: 100%;
   z-index: 10001;
+  @media screen and (min-width: 1024px) {
```
mathio, line 42: Did you test this locally? This does not seem to fix the issue with centering it. The min-width and min-height are specially needed for small screens (under 1024).

mathio: I think you need to change the logic in `popupWrapper` below where `top` and `left` are calculated.

antoniofull: 🤔  I see it centered, also the ticket talks about desktop not mobile 😕

mathio: Did you check if it is really centered?

Screenshot 2020-11-23 at 16 00 38

Since we do not determine mobile based on screen size, I think the popup should keep minimum size on desktop regardless of screen size - that means even for screens < 1024px wide.

antoniofull: Are we looking at different embeds? 
Screenshot 2020-11-23 at 16 13 53

antoniofull: > I think the popup should keep minimum size on desktop regardless of screen size - that means even for screens < 1024px wide

Sorry mate, this line is quite confusing to me,  we talk about desktop then < than 1024px 😄  shall we jump in a chat?

src/core/views/popup.js
```diff
@@ -56,25 +60,18 @@ const Overlay = styled.div`
...
-    return `
-      width: calc(${p.size}% - 80px);
-      height: calc(${p.size}% - 80px);
-      top: calc(${offset}% + 40px);
-      left: calc(${offset}% + 40px);
-    `
-  }}
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);\${({})}
```
mathio, line 73: What is `\${({})}`?

src/core/views/popup.js
```diff
@@ -56,25 +60,18 @@ const Overlay = styled.div`
...
+  ${({ size }) => size && dimensions}
```
mathio, line 74: Great job removing that complicated calculations!

One minor thing - I would not keep the `dimensions` in separate variable since its not reused anywhere.

src/core/views/popup.js
antoniofull: leftover from the rebase :(

antoniofull: I wonder why it did not fail? (magic of css 🤣 )

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/06d8ab31e853fd7ff8d8fc76c237bbfe6615ac9c/embed.js

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/b3042fd5af12ebdeef819d2bfc242d8b3052d1f3/embed.js

src/core/views/popup.js
mathio: I would remove the media query and I think its good to go. With media query this is what it looks like on desktop when you resize browser window: 
Screenshot 2020-11-24 at 17 15 11

antoniofull: You're right, I messed up when rebasing from master (I had conflicts)

mathio (approved):
✅

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/084b4e659db88af8ec294bf19f3c63e3962e8561/embed.js

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/5c1d4143632851a4c45f6ef1ccf916058f481da8/embed.js

PR #108, opened 2020-11-25
PI-89 Add shareGoogleAnalyticsInstance option feature

supertinou:
# Motivation

Support the improvement to the Google Analytics Integration so that the Google analytics user session on the website where the Typeform is embedded is share to Typeform and even beyond and so that as the owner of the website I can get the accurate full view of the experience of the user and can get to improve it.

This specific feature is part of the bigger epic: https://jira.typeform.tf/browse/PI-88

# Under the hood

The changes allow you to specify a ```shareGoogleAnalyticsInstance```  option in the embed
and this had the effect to pass in a ```shareGoogleAnalyticsInstance``` parameter to the Typeform URL and to, if there is Google analytics in the page of the embed to pass a ```ga-client-id``` parameter via ```postMessage``` as soon as the google Analytics parameter Id is available.

# How to use the feature.

1) Enable the Google analytics integration

2) Enable the ```shareGoogleAnalyticsInstance``` in the embed:

In an HTML snippet using the embed:

```html
<div
      class="typeform-widget"
      data-url="mytypeformURL"
      data-share-google-analytics-instance
    ></div>
<script type="text/javascript" src="https://embed.typeform.com/embed.js"></script>
```

Using the API: 

```js

const embedElement = document.querySelector('.js-typeform-embed')

typeformEmbed.makeWidget(
  embedElement,
  'https://admin.typeform.com/to/PlBzgL',
  {
    buttonText: "Answer this!",
    shareGoogleAnalyticsInstance
  }
)
```

# JIRA: https://jira.typeform.tf/browse/PI-89

# Related PR: https://github.com/Typeform/louvre/pull/1330


mathio (commented):
Please make sure the build is passing. If you need any help let us know.

antoniofull:
@supertinou I am going to update this branch as it is outdated (and trigger visual tests again). I don't think you have access to the visual tests, in case ping me if you need help with fixing those tests 😄

supertinou:
@antoniofull, I'd love a second pair of eyes if you get the chance, regarding the ```> this.setState is not a function``` preventing the visual test to pass. I haven't been able to figure out why this error is being raised yet 🤔 🙂.

mathio:
@supertinou is that happening only for visual tests or can you replicate the issue locally? Most common cause for such issue is that you are calling `this.setState(...)` in a function method you did not bind:

```
this.myFunction = this.myFunction.bind(this)
```

src/core/views/mobile-modal.js
```diff
@@ -51,6 +51,7 @@ class MobileModal extends Component {
     }
 
     this.handleMessage = this.handleMessage.bind(this)
+    this.handleFormReady = callIfEmbedIdMatches(this.handleFormReady.bind(this), this.props.embedId)
```
supertinou, line 54: And that was it, thank you :)

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/c271ea0ac0cd051eb0463937026d88ef22f6a080/embed.js

supertinou:
I've been playing around to add unit tests to freeze internally the googleAnalyticsInstanceSharing in `views/mobile-modal.js`, `views/popup.js` and `views/widget.js` but could not find a way to either:
- mock the `google-analytics-instance-sharing.js` function to make sure it is called internally when `share-google-analytics-instance` is enabled.
- add a google analytics tag in a unit test and check that a postMessage event has been sent.
If you have any suggestion, let me know :)

src/core/features/google-analytics-instance-sharing.js
```diff
@@ -0,0 +1,24 @@
+const setupGoogleAnalyticsInstanceSharingFeature = (iframeRef, embedId) => {
+  // Throw an error if the feature is enabled but ga is not found
+
+  const gaObject = window[window.GoogleAnalyticsObject]
+
+  if (gaObject === undefined || gaObject === null) {
+    throw new Error(
```
lfons, line 7: Should we use a try-catch block to better manage this function call? What happens when an error is thrown (E.g. shareGoogleAnalyticsInstance is set to true, but no gaObject is found)? Is the embedded form still being displayed?

lfons (changes requested):

lfons (commented):
It seems that we can set the GA code for the form thought the connect panel, how does this play with setting the GA code from the embed widget? What happens if the form already has a GA code, and on the other side, we are passing a new one from the embed (from the parent page containing an embed as per the PR implementation)?

supertinou (commented):
Regarding: 
> It seems that we can set the GA code for the form thought the connect panel, how does this play with setting the GA code from the embed widget? What happens if the form already has a GA code, and on the other side, we are passing a new one from the embed (from the parent page containing an embed as per the PR implementation)?

Thank you for reviewing with attention and care :)

Exactly, we can have the scenario that the `GA code` on the website where the form is embedded is different than the `GA code` set for the integration in the connect panel. 

In this scenario, the embed will send a postMessage `ga-client-id` event to the iframe (louvre) with a `gaClientId` which is associated to a different `GA code` than the one for which we is set in the connect panel.

I just tested it in the subscribe environment:

Then the end results are that in the Google Analytics dashboard for both `GA code` / GA account which are totally separate, there will be a similar `gaClientId` in the list of users in "User explorer" with the events from the website in the `GA code` (1) and then the events sent by the Typeform in `GA Code` (2)

So... the user needs to make sure that they set similar `GA code` for their website and the Typeform they embed.

src/core/features/google-analytics-instance-sharing.js
lfons: Thanks for getting back to me @supertinou 

IMHO Ideally, it should work as you said because it would be self-evident that something is wrong. 

However, to always display the form if we can, I will display an error in the console BUT continue with the execution and display the Typeform. 

For example, the GA code may be removed from the page because they want to use other analytics tools, then the form will stop working. Also, sometimes, non-technical people may be in charge of setting these GA codes (through Google Tag Manager, for example), they can accidentally remove the GA code, and then the form will stop working. 

So, to be on the safe side, I will display an error in the console but still show the form in these cases.

lfons:
> Regarding:
> 
> > It seems that we can set the GA code for the form thought the connect panel, how does this play with setting the GA code from the embed widget? What happens if the form already has a GA code, and on the other side, we are passing a new one from the embed (from the parent page containing an embed as per the PR implementation)?
> 
> Thank you for reviewing with attention and care :)
> 
> Exactly, we can have the scenario that the `GA code` on the website where the form is embedded is different than the `GA code` set for the integration in the connect panel.
> 
> In this scenario, the embed will send a postMessage `ga-client-id` event to the iframe (louvre) with a `gaClientId` which is associated to a different `GA code` than the one for which we is set in the connect panel.
> 
> I just tested it in the subscribe environment:
> 
> Then the end results are that in the Google Analytics dashboard for both `GA code` / GA account which are totally separate, there will be a similar `gaClientId` in the list of users in "User explorer" with the events from the website in the `GA code` (1) and then the events sent by the Typeform in `GA Code` (2)
> 
> So... the user needs to make sure that they set similar `GA code` for their website and the Typeform they embed.

Thanks for your explanation @supertinou. Has this behavior been validated with Product? If yes, and this is what we want, I am happy with it :)

src/core/features/google-analytics-instance-sharing.js
supertinou: Yes, we can keep the error in the console and the user will notice in their Google Analytics Dashboard if they have non matching users. Adding that to the integration for potential errors for the CS team!

supertinou: I addressed changes here: https://github.com/Typeform/embed/pull/108/commits/63b4b9847ed43ae5ea3b37a9dd7abd4477266a0a :)

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/8721356547fbda570a14715365c60f70072c5cd5/embed.js

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/5bd9dbf23fb23954d94743f10a06c9ee293412fa/embed.js

lfons (approved):

mathio (approved):

PR #110, opened 2020-12-02
feat(DIST-401): Add tracking parameters

lfons:
Adds the following metadata options and forwards them to renderer requests:

- source field as typeform-source query parameter. 
- medium field as typeform-medium query parameter. If empty set to embed-sdk
- mediumVersion field as typeform-medium-version query parameter

Uses existing open option and adds it as typeform-embed-trigger-type URL parameter.

- [x]  Link to the ticket [https://jira.typeform.tf/browse/DIST-401]()

### DIFF
--- a/src/core/attributes.js
+++ b/src/core/attributes.js
@@ -35,8 +35,42 @@ const getDataset = element => {
   return data
 }
 
-const sanitizePopupAttributes = data => {
+const sanitizeCommonAttributes = data => {
   const obj = {}
+
+  if (data.hideHeaders === '' || data.hideHeaders === 'true') {
+    obj.hideHeaders = true
+  }
+
+  if (data.hideFooter === '' || data.hideFooter === 'true') {
+    obj.hideFooter = true
+  }
+
+  if (data.hideScrollbars === '' || data.hideScrollbars === 'true') {
+    obj.hideScrollbars = true
+  }
+
+  if (data.source) {
+    obj.source = data.source
+  }
+
+  if (data.medium) {
+    obj.medium = data.medium
+  }
+
+  if (data.mediumVersion) {
+    obj.mediumVersion = data.mediumVersion
+  }
+
+  if (data.embedTriggerType) {
+    obj.embedTriggerType = data.embedTriggerType
+  }
+
+  return obj
+}
+
+const sanitizePopupAttributes = data => {
+  const obj = sanitizeCommonAttributes(data)
 
   if (data.mode) {
     obj.mode = transformLegacyDataMode(data.mode)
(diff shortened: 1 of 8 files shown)


travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/f4d18b36a42d5f7544985013f3b66aec3b673e10/embed.js

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/cf5877328a913df81bf15aae8c0958d582e92532/embed.js

src/core/attributes.js
```diff
@@ -35,9 +35,43 @@ const getDataset = element => {
...
+  if (data.mediumVersion) {
+    obj.mediumVersion = data.mediumVersion
+  }
+
+  if (data.embedTriggerType) {
+    obj.embedTriggerType = data.embedTriggerType
+  }
+
+  return obj
+}
```
antoniofull: would a switch here make it more easy to read and maintain?

antoniofull (approved):

antoniofull:
LGTM

src/core/make-widget.js
```diff
@@ -30,6 +31,9 @@ const defaultOptions = {
 
 const queryStringKeys = {
   mode: 'typeform-embed',
+  source: 'typeform-source',
+  medium: 'typeform-medium',
+  mediumVersion: 'typeform-medium-version',
```
mathio, line 36: From [the ticket](https://jira.typeform.tf/browse/DIST-401):

> mediumVersion  ... If medium is embed-sdk, set to **current embed sdk version**, otherwise set to **empty string**

I think you need to set the mediumVersion value to current embed library version. The ticket does not say it can be overridden, but I guess it's fine.

src/core/make-widget.js
```diff
@@ -30,6 +31,9 @@ const defaultOptions = {
 
 const queryStringKeys = {
   mode: 'typeform-embed',
+  source: 'typeform-source',
```
mathio, line 34: From [the ticket](https://jira.typeform.tf/browse/DIST-401):

> Source: Optional value. **Domain name of the site using the SDK.**

We should read the domain from `window.location` if it is not set.

src/core/attributes.js
mathio: Its not if-else, I dont think we can use `switch` here and keep the same functionality.

src/core/make-widget.js
lfons: To my understanding, the mediumVersion is the version of whatever medium we are using. E.g. Embed-skd, embed-snippet, embed-wordpress-plugin, embed-hotspot-plugin, etc. So it is not necessary always the embed-sdk library version. 

However, if we have to default to embed-sdk, because medium and/or mediumVersion are not provided, then yes, we could use the embed library version. I guess we will need to do some versioning work (e.g. how do we get the actual version number in our code?), and this should probably be implemented in another ticket.

src/core/make-widget.js
lfons: Good catch Matej. Implementing this one ASAP.

src/core/make-widget.js
mathio: > if we have to default to embed-sdk, because medium and/or mediumVersion are not provided, then yes, we could use the embed library version

Thats correct. Is there a ticket for that? From ticket description my understanding is it should be part of this ticket.

But I think we can not read current lib version right now. @prilutskiy do you have any ideas on how to access it?

travis-tf:
*[BOT]* Preview available: https://embed.typeform.com/c9e57ff98648239e4b5bfebf0d39c58f645701b0/embed.js

mathio (approved):

PR #111, opened 2020-12-09
Update Opacity in README.md

jeffshek:
Setting opacity of 0 is a bit surprising, background images won't show up as a result. When debugging, I was thinking it was a CSS issue, or a https (testing from my local environment).


mathio (approved):

mathio:
Thank you for your PR @jeffshek

PR #133, opened 2021-01-27
fix(DIST-629): Update iframe reload logic

mathio:
When widget is opened as popup on mobile the iframe needs to be reloaded to make sure we display welcome screen again after the popup is closed. Our current logic stopped working.

https://jira.typeform.tf/browse/DIST-629

Widget with issue: https://embed-typeform.glitch.me/
Fixed widget (preview): https://embed.typeform.com/f647bf19c5aaccc56feaa25c397c117cfb8c0714/widget-real.html

Note: The issue is apparent only on mobile when widget is opened in popup.

### DIFF
--- a/src/core/views/widget.js
+++ b/src/core/views/widget.js
@@ -57,20 +57,11 @@ const PlaceholderAnimationAppear = keyframes`
 `
 
 const PlaceholderAnimationDisappear = keyframes`
-  100% {
-    opacity: 0;
-  }
-
-  75% {
-    opacity: 1;
-  }
-
-
-  25% {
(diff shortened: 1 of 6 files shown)


src/core/views/widget.js
```diff
@@ -57,20 +57,11 @@ const PlaceholderAnimationAppear = keyframes`
 `
 
 const PlaceholderAnimationDisappear = keyframes`
```
mathio, line 59: This removes a flash after popup is closed.

mrisoli (approved):
🥩

PR #135, opened 2021-01-29
Stale bot

prilutskiy:


.github/stale.yml
```diff
@@ -0,0 +1,60 @@
...
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 2
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
```
mathio, line 14: Maybe we should add `dependencies` or this will start closing dependabot PRs. Do we want that? We are not that fast at merging those.

.github/stale.yml
```diff
@@ -0,0 +1,60 @@
...
+  - pinned
+  - security
+  - '[Status] Maybe Later'
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
```
mathio, line 20: Does `false` mean this will also close stale issues? We have some issues that have been around for quite long and it does not look like we are ready to close them.

.github/stale.yml
prilutskiy: Good point!

.github/stale.yml
prilutskiy: let's use a new label `backlog` for such cases. WDYT?

mathio: In that case should we exclude issues with `backlog` tag too? Or do you want to close those?

prilutskiy: yeah, I've done exactly that, added `backlog` and `dependencies` labels to that list.

mathio (approved):

PR #137, opened 2021-02-02
test(DIST-711): Add jest support

mathio:
Add unit tests for existing code.

Merge #136 first

### DIFF
--- a/packages/embed/package.json
+++ b/packages/embed/package.json
@@ -11,6 +11,7 @@
   "scripts": {
     "build": "tsc",
     "dev": "tsc -w",
-    "lint": "eslint src --ext .js,.ts,.jsx,.tsx --max-warnings=0"
+    "lint": "eslint src --ext .js,.ts,.jsx,.tsx --max-warnings=0",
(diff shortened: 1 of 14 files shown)


packages/embed/package.json
```diff
@@ -5,9 +5,18 @@
...
+  },
   "scripts": {
     "build": "tsc",
     "dev": "tsc -w",
-    "lint": "eslint src --ext .js,.ts,.jsx,.tsx --max-warnings=0"
+    "lint": "eslint src --ext .js,.ts,.jsx,.tsx --max-warnings=0",
+    "test": "jest"
+  },
+  "devDependencies": {
+    "@types/jest": "^26.0.20",
```
mathio, line 18: Should we put dependencies here or to the root? How to do this? I am not familiar with lerna. @prilutskiy please advise.

prilutskiy: Let's keep everything in the root, so the same packages' versions could be reused across all projects

prilutskiy (approved):

PR #140, opened 2021-02-03
feat: Add prettier

mathio:
Eslint already checks prettier syntax for JS and TS files, this will check for other files too.


mathio:
Wrong branch :)

PR #146, opened 2021-02-04
fix(DIST-713): Force iframe redraw onLoad

prilutskiy:


mathio (approved):

src/core/views/components/iframe.js
```diff
@@ -20,8 +21,32 @@ class Iframe extends Component {
     this.iframeRef = node
   }
 
-  handleLoad() {
-    this.props.onLoad && this.props.onLoad(this.iframeRef)
+  async handleLoad() {
+    this.props.onLoad(this.iframeRef)
+    await this.triggerIframeRedraw()
```
scarciofolomarco, line 26: won't `this.forceUpdate()` just work for that?

prilutskiy: It's about asking _**the browser**_ to redraw the iframe, rather than asking react to rerender

PR #152, opened 2021-02-08
chore: add webpack for embed browser build

mrisoli:
## Description

Add webpack, webpack-cli and ts-loadr as devDependencies for embed
Add webpack config for browser bundle
Add script to package.json to build with webpack

## Checklist

### DIFF
--- a/packages/embed/package.json
+++ b/packages/embed/package.json
@@ -23,6 +23,7 @@
     "test:functional": "start-server-and-test demo 6006 cy:run"
   },
   "devDependencies": {
+    "html-webpack-plugin": "^5.0.0",
     "ts-loader": "^8.0.15",
     "webpack": "^5.21.2",
     "webpack-cli": "^4.5.0"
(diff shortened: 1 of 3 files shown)


packages/embed/package.json
```diff
@@ -3,13 +3,15 @@
   "version": "1.0.0",
   "private": true,
   "main": "build/index.js",
+  "browser": "dist/webpack/bundle.js",
```
prilutskiy, line 6: I'm curious, how does this work?

prilutskiy (approved):

packages/embed/package.json
mathio: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#browser

mrisoli: This is just a hint field, it doesn't directly do anything but hint that file is supposed to be used in the browser.

prilutskiy: Oh, okay, I thought it maybe automatically does something :)

mathio (approved):

PR #153, opened 2021-02-08
feat(DIST-709): Add popup functionality

mathio:


packages/embed/src/factories/create-popup/create-popup.ts
```diff
@@ -10,17 +10,43 @@ export type Popup = {
...
-    open: () => popup.classList.add('typeform-popup-opened'),
-    close: () => popup.classList.remove('typeform-popup-opened'),
-    refresh: () => iframe.contentWindow?.location.reload(),
+    open,
+    close,
+    refresh,
   }
 }
+
+export const buildPopup = (iframe: HTMLIFrameElement): HTMLDivElement => {
```
mathio, line 39: I prefer this here in this file while its short. If it becomes very long we can extract it to separate file. WDYT?

mathio: On the other hand we would need to do `export { createPopup } from './create-popup'` in `./index.ts` to make sure we dont export this to library consumers. I am open to ideas.

mrisoli (approved):

packages/embed/src/factories/create-popup/create-popup.ts
prilutskiy: My first thought was - why even exporting? - and I see you need it for unit tests.

Well, then we can place these helper functions like this:
```
create-popup/
  utils/
    build-popup.ts
    build-close-button.ts
    index.ts
  create-popup.ts
  create-popup.spec.ts
  index.ts
```
so we both export the functions from their source files for testing, and do not export them outside the create-popup.

mathio: Good idea. I will name the directory `elements` not `utils`, since anything can be "utils".

PR #156, opened 2021-02-08
feat: Use form ID instead of whole URL

mathio:
We dont need whole URL, since we recommend "form" subdomain to avoid possible issues
when subdomain is changed. Standard subdomain "form" will always work.


prilutskiy (approved):
Awesome 😍

PR #158, opened 2021-02-09
test(DIST-693): Add Applitools and visual tests

antoniofull:
[Ticket DIST-693](https://jira.typeform.tf/browse/DIST-693)
- Adds applitools basic config
- I don't know yet how we want to run these tests, via storybook?
- I am not sure I added the files and the scripts in the correct paths ( I followed our cypress config in `packages/embed`), so please review. 
- [x] Link to the issue - if any [DIST-693](https://jira.typeform.tf/browse/DIST-693)


prilutskiy (approved):

package.json
```diff
@@ -8,7 +8,9 @@
   "workspaces": [
     "packages/*"
   ],
+
```
mathio, line 11: There shouldnt be an empty line in `package.json`.

packages/embed/e2e/applitools-utils.js
```diff
@@ -0,0 +1,22 @@
+import { screenSizeDesktop, screenSizeMobile } from '../../../e2e/cypress-utils'
+
+const eyesCheck = async (name, mobile) => {
```
mathio, line 3: This should be a `.ts` file and all functions need to be typed.

packages/embed/e2e/applitools-utils.js
```diff
@@ -0,0 +1,22 @@
+import { screenSizeDesktop, screenSizeMobile } from '../../../e2e/cypress-utils'
```
mathio, line 1: This import is not resolving correctly for me. It should be `./cypress-utils` right?

packages/embed/e2e/cypress-utils.js
```diff
@@ -0,0 +1,31 @@
+export const IFRAME_SELECTOR = '[data-qa="iframe"]'
+export const screenSizeDesktop = { width: 1024, height: 768 }
+export const screenSizeMobile = { width: 375, height: 667 }
+
+const setViewport = ({ width, height }) => {
+  cy.viewport(width, height)
+}
+
+export const getIframeBody = (iframe = IFRAME_SELECTOR) => {
```
mathio, line 9: Is this used anywhere? I would like to avoid this, we should not test <iframe> body anyway, but rather its attributes (eg. `src`). What is inside an iframe is displayed by renderer.

Also, typescript for this file too.

packages/embed/package.json
```diff
@@ -20,7 +20,9 @@
     "test": "jest",
     "cy:run": "yarn cypress run --browser chrome --headless",
     "cy:open": "yarn cypress open",
-    "test:functional": "start-server-and-test demo 6006 cy:run"
+    "cy:visual": "yarn cypress run --spec e2e/spec/visual/**/* --browser chrome",
```
mathio, line 23: - Lets rename `cy:run` to `cy:functional` and add `--spec` param as well to make sure it runs only functional tests.
- You also need to move functional spec to its own dir.
- We need to run visual tests in Github Actions on PR.
- Please write at least one test for demonstration purposes. You can open stories directly without the storybook UI like we do for functional tests now:
```
cy.visit('/iframe.html?id=embed-popup--popup&viewMode=story')
```

mathio (changes requested):
Let me know on slack me if you have any questions.

packages/embed/e2e/applitools-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    },
+    batchName: 'Embed',
+  })
+  cy.eyesCheckWindow(name)
+  cy.eyesClose()
+}
+
+export const eyesCheckDesktop = (name: string) => eyesCheck(name, false)
+
+export const eyesCheckMobile = (name: string) => eyesCheck(name, true)
```
mathio, line 22: Lets make this a bit more readable:

```
import { Viewport } from './cypress-utils.ts'

const eyesCheck = async (name: string, { width, height }: Viewport) => { ... }

export const eyesCheckDesktop = (name: string) => eyesCheck(name, screenSizeDesktop)

export const eyesCheckMobile = (name: string) => eyesCheck(name, screenSizeMobile)
```

WDYT?

.github/workflows/pull-request-next.yml
```diff
@@ -37,3 +41,20 @@ jobs:
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
       - run: yarn install
       - run: yarn lerna run test:functional
+  visual:
+    name: Visual Tests
```
mathio, line 45: Other jobs do not have names (they use job id as name, eg `functional`). Please remove.

mathio (approved):

packages/embed/e2e/applitools-utils.ts
antoniofull: sounds good

PR #159, opened 2021-02-09
feat(DIST-706): Add widget functionality

mathio:


packages/embed/src/factories/create-widget/elements/build-widget.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { buildWidget } from './build-widget'
+
+describe('build-widget', () => {
+  describe('#buildWidget', () => {
```
mrisoli, line 4: What's up with this context nesting? Is It filename->function?

mrisoli (approved):

packages/embed/src/factories/create-widget/elements/build-widget.spec.ts
mathio: Correct. This file has just one function, but it could have more if necessary.

prilutskiy (approved):

PR #160, opened 2021-02-09
feat(DIST-707): Remove full page type

mathio:
Same functionality can be achieved by a widget in full screen container.
We can add shorthand once widget is fully implemented, if necessary.

Also, adding widgets to end of body does not really make sense. It breaks single page apps (body is never reloaded) and on script reload it adds multiple forms.

### DIFF
--- a/packages/embed/src/base/embed-types.ts
+++ b/packages/embed/src/base/embed-types.ts
@@ -1,1 +1,1 @@
-export type EmbedType = 'fullpage' | 'widget' | 'popup' | 'drawer_left' | 'drawer_right' | 'popover' | 'side_panel'
+export type EmbedType = 'widget' | 'popup' | 'slider' | 'popover' | 'side-tab'
--- a/packages/embed/src/factories/create-widget/create-widget.ts
+++ b/packages/embed/src/factories/create-widget/create-widget.ts
@@ -11,8 +11,7 @@ export const createWidget = (formId: string, options: WidgetOptions): Widget => 
   const iframe = createIframe(formId, 'widget', options)
   const widget = buildWidget(iframe)
 
-  const container = options.container || document.body
-  container.append(widget)
+  options.container.append(widget)
 
   return {
     refresh: () => iframe.contentWindow?.location.reload(),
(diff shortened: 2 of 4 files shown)


prilutskiy (approved):
How about having a shortcut? So the API is clear and consistent with our share page?

mathio:
Yes, we can make shortcuts once the widget is fully implemented.

The main difference in use cases:
* **share page** - I want to create a new HTML page and my HTML knowledge is limited, I want to copy&paste the code including whole page markup `<!DOCTYPE html><html><head>...</html>` and display fullscreen form on a page
* **embed lib** - I already have a page with some markup and I want to add typeform, if I want to make it full page I most likely want to keep header and footer or other navigation

prilutskiy:
> Yes, we can make shortcuts once the widget is fully implemented.
> 
> The main difference in use cases:
> 
> * **share page** - I want to create a new HTML page and my HTML knowledge is limited, I want to copy&paste the code including whole page markup `<!DOCTYPE html><html><head>...</html>` and display fullscreen form on a page
> * **embed lib** - I already have a page with some markup and I want to add typeform, if I want to make it full page I most likely want to keep header and footer or other navigation

I believe there's a way to generate and use something like this:
```html
<div class="typeform-fullpage"></div>
<script src="xxx"></script>
```

instead of `<!DOCTYPE html><html><head>...</html>`. Browsers add missing HTML / head / body automatically, and we can leverage that providing a user just with something human readable.

mathio:
I dont know how recent is [this MDN article](https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode) but we might need to use DOCTYPE to make sure browsers render content as expected. Also, meta tag with encoding might be necessary too.

We might provide a pre-formatted code if we want it readable :)
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset=UTF-8>
    <title>Typeform Title from Builder</title>
  </head>
  <body>
    <div class="typeform-fullpage"></div>
    <script src="xxx"></script>
  </body>
</html>
```

PR #162, opened 2021-02-10
chore(DIST-690): Use semantic-release-monorepo to deploy packages

mrisoli:
## Description

This PR adds semantic-release packages with semantic-release-monorepo to use lerna to deploy packages from next branch.

## Checklist


prilutskiy:
How about the `.releaserc` config file? Don't we need it?

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+          token: ${{ secrets.GH_TOKEN }}
+      - uses: actions/setup-node@v1
+      - uses: actions/cache@v2
+        with:
+          path: |
+            **/node_modules
+            ~/.cache
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - run: yarn install
+      - run: yarn build
```
mathio, line 48: Maybe we should run the build via lerna to build each package?

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+      # Update registry and tokens with write access for releasing.
+      - run: rm ./.npmrc
+      - run: npm config set '//registry.npmjs.org/:_authToken' $NPM_TOKEN
+      - run: npm config set '//npm.pkg.github.com/:_authToken' $GH_TOKEN
+      - run: npm config set @typeform:registry https://npm.pkg.github.com/
+      - run: yarn run lerna exec --concurrency 1 npx --no-install -- -- semantic-release
```
mathio, line 54: `yarn lerna run semantic-release` does not work? Or do we need the concurrency  to be 1?

I would even suggest to run `yarn lerna run release` and implement `release` script in each package. For now it will be just `"release": "yarn semantic-release",` but later we can have customized release processes for each package.

mathio (approved):
I approve. Please merge tomorrow since this will release a new version. It [should release as "next"](https://semantic-release.gitbook.io/semantic-release/v/beta/usage/configuration, but if something goes wrong we will be available (eg. if we make an incorrect release and need to revert).

.github/workflows/release.yml
mrisoli: Maybe it would, I took it from here:
https://github.com/pmowrer/semantic-release-monorepo#with-lerna
I'm assuming having some concurrency might be interesting for building multiple packages

.github/workflows/release.yml
mrisoli: So, replace it with `yarn lerna bootstrap`?

mathio: Like this maybe?
```
yarn lerna bootstrap
yarn lerna run build
```

PR workflow does  `yarn install` only. Maybe it should be updated too.

antoniofull (approved):

.releaserc
```diff
@@ -0,0 +1,23 @@
+{
+  "branches": [
+    "master",
```
prilutskiy, line 3: let's remove master branch from this config, just in case

prilutskiy (approved):

PR #167, opened 2021-02-15
fix(DIST-730): Remove display toggle that causes flicker

mrisoli:

### DIFF
--- a/src/core/views/components/iframe.js
+++ b/src/core/views/components/iframe.js
@@ -36,13 +36,9 @@ class Iframe extends Component {
         resolve()
         return
       }
-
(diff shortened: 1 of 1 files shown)


prilutskiy:
please apply the same changes to the next branch, `packages/embed/src/utils/create-iframe/trigger-iframe-redraw.ts`

prilutskiy (changes requested):
please apply the same changes to the next branch, `packages/embed/src/utils/create-iframe/trigger-iframe-redraw.ts`

prilutskiy (approved):

PR #171, opened 2021-02-18
Initialize embeds from HTML attributes, auto-inject CSS, add demo packages

mathio:
The PR is split into multiple commits:

* chore: Remove kitchen sink package 
  _Kitchen sink does not implement the library using a real world approach. It abstracts some concepts, requires additional maintenance and could deceive developers into thinking it works while real-world implementation does not._
* [feat: Initialize embed based on HTML attributes and auto-inject CSS](https://github.com/Typeform/embed/pull/171/commits/b92d602426321d6a61b4a78a6e6f912880573a22)
  * builds SCSS into CSS and exposes it publicly
  * when the lib is included as JS file via `<script>`
    * introduces `src/initializers/*` to insert embeds based on `data-tf-*` attributes
    * injects CSS when embed is created via `data-tf-*`
  * when lib is imported from `node_modules` it is up to consumer to link to CSS (or import it and bundle it with their own CSS)
* feat: Add demo for static HTML sites
* feat: Add demo for sites served via webpack
* feat: Add demo for react apps
* test: Update functional tests to use demo-html
* chore: Install dependencies in CI with --frozen-lockfile
  _It signifficantly speeds up the build._


packages/embed/src/browser.ts
```diff
@@ -0,0 +1,12 @@
+import * as initializers from './initializers'
```
prilutskiy, line 1: nice!

packages/embed/src/factories/create-popup/create-popup.ts
```diff
@@ -6,10 +6,11 @@ import { buildCloseButton, buildPopup } from './elements'
 export type Popup = {
   open: () => void
   close: () => void
+  toggle: () => void
```
prilutskiy, line 9: nice shortcut. Let's add a unit test

packages/embed/src/initializers/initialize-popups.ts
```diff
@@ -0,0 +1,21 @@
+import { createPopup } from '../factories/create-popup'
+import { includeCss } from '../utils'
+
+const POPUP_ATTRIBUTE = 'data-tf-popup'
```
prilutskiy, line 4: let's move it to `src/constants` to avoid duplicates in future

packages/embed/webpack.config.js
```diff
@@ -1,43 +1,56 @@
-const HtmlWebpackPlugin = require('html-webpack-plugin')
 const path = require('path')
 
-const npmConfig = {
+const sass = require('node-sass')
+const CopyPlugin = require('copy-webpack-plugin')
+
+const baseConfig = {
   mode: 'development',
```
prilutskiy, line 7: how about `mode: process.env.NODE_ENV` ?

prilutskiy (approved):

mrisoli (approved):

PR #173, opened 2021-02-19
onReady, onSubmit, onScreenChanged

prilutskiy:


mathio (commented):
Do you think we could also pass [functions as element attributes](https://stackoverflow.com/questions/32795225/pass-a-callback-function-as-a-html-data-attribute)?

mathio (commented):
Also, I think this would be a great candidate for functional tests. I can help if you want.

packages/embed/src/utils/load-options-from-attributes.ts
```diff
@@ -25,6 +25,11 @@ const transformInteger = (value: string | null): number | undefined => {
   return isNaN(integer) ? undefined : integer
 }
 
+const transformFunction = (value: string | null): Function | undefined => {
+  const fn = value && value in window ? window[value] : undefined
```
mathio, line 29: Does this work? When I played with this I had to do `window[value].call()`.

packages/embed/src/utils/build-iframe.src.spec.ts
```diff
@@ -3,21 +3,26 @@ import { buildIframeSrc } from './build-iframe-src'
...
       )
     })
 
     it('should include url options', () => {
-      expect(buildIframeSrc('some-id', 'widget', { source: 'unit-test-source' })).toBe(
-        'https://form.typeform.com/to/some-id?typeform-embed=embed-widget&typeform-source=unit-test-source'
-      )
+      expect(
+        buildIframeSrc({ formId: 'some-id', type: 'widget', embedId: '', options: { source: 'unit-test-source' } })
+      ).toBe('https://form.typeform.com/to/some-id?typeform-embed=embed-widget&typeform-source=unit-test-source')
```
mathio, line 14: Do we have a different prettier settings?  🤔

prilutskiy: why?

packages/embed/src/utils/load-options-from-attributes.ts
prilutskiy: isn't it supposed to return a function, instead of calling it? If so - yep, it works, also there's a unit test you can check out

packages/embed/src/utils/create-iframe/get-form-event-handler.ts
```diff
@@ -0,0 +1,33 @@
...
+) {
+  return (event: any) => {
+    if (event.data.type !== eventType) {
+      return
+    }
+    if (event.data.embedId !== embedId) {
+      return
+    }
+    const transformedEvent = transformEventPayload(event)
+    callback?.(transformedEvent)
```
mathio, line 31: I would prefer to be more explicit:

```
if (!callback) {
  return
}
```

packages/embed/src/utils/load-options-from-attributes.ts
mathio: You are right. In that case should it return `window[value].call`? The `window[value]` is not callable with just `()`.

packages/embed/src/utils/build-iframe.src.spec.ts
mathio: Oh, you added `'embedId'`. I overlooked it, sorry :)

packages/embed/src/utils/create-iframe/get-form-event-handler.ts
prilutskiy: isn't this why the operator even exists? 🙂

mathio: I think it kind of hides the functionality but that might be personal preference. Also, with my suggestion you would skip `transformEventPayload` when callback is not defined :)

packages/embed/src/utils/load-options-from-attributes.ts
prilutskiy: Why do you think it shouldn't work?

prilutskiy: also, .call signature is not the same as the function's signature

prilutskiy: image

For instance, it works with `window.getSelection`. Or you have some edge case in mind?

packages/embed/src/utils/create-iframe/get-form-event-handler.ts
prilutskiy: `transformEventPayload` falls back to `(ev) => undefined` when it's not defined, meaning that if we don't expose anything explicitly, it's not exposed automatically. The fallback syntax is not the best there, it's easy to overlook :)

antoniofull (approved):

packages/embed/src/utils/load-options-from-attributes.ts
mathio: I double checked and now it works for me too :) 👍

PR #176, opened 2021-02-22
feat(DIST-700): Add transferableUrls parameters

antoniofull:
Adds transferableURL parameters
- [x] Link to the issue - if any [DIST-700](https://jira.typeform.tf/browse/DIST-700)

### DIFF
--- a/packages/embed/src/utils/url-parameters-transfer.ts
+++ b/packages/embed/src/utils/url-parameters-transfer.ts
@@ -1,0 +1,32 @@
+type Params = { [key: string]: string }
+
+type QueryStrings = {
+  [key: string]: any
+}
+
+const urlSearchToParams = (search: string) => {
+  var params: Params = {}
+  if (search !== '' && search !== null) {
+    var vars = search.split('&')
+    for (var i = 0; i < vars.length; i++) {
+      var pair = vars[i].split('=')
+      params[pair[0] as keyof Params] = decodeURIComponent(pair[1])
+    }
+  }
+  return params
+}
+
+export const transferUrlParametersToQueryStrings = (
+  transferableUrlParameters: string[],
+  queryStrings: QueryStrings
+) => {
+  const urlSearchString = window.location.search.substr(1)
+  const urlSearchParams = urlSearchToParams(urlSearchString)
+  const queryStringsWithTransferedParams = { ...queryStrings }
+  transferableUrlParameters.forEach((transferableParam: string) => {
+    if (!(transferableParam in queryStrings)) {
+      queryStringsWithTransferedParams[transferableParam] = urlSearchParams[transferableParam]
+    }
+  })
+  return queryStringsWithTransferedParams
+}
--- a/packages/embed/src/utils/url-parameters-transfer.spec.ts
+++ b/packages/embed/src/utils/url-parameters-transfer.spec.ts
@@ -1,0 +1,44 @@
+import { transferUrlParametersToQueryStrings } from './url-parameters-transfer'
+
+describe('transferUrlParametersToQueryStrings', () => {
+  const location: Location = window.location
+
+  beforeAll(() => {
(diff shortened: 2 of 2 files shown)


packages/embed/src/utils/url-parameters-transfer.ts
```diff
@@ -0,0 +1,32 @@
...
+) => {
+  const urlSearchString = window.location.search.substr(1)
+  const urlSearchParams = urlSearchToParams(urlSearchString)
+  const queryStringsWithTransferedParams = { ...queryStrings }
+  transferableUrlParameters.forEach((transferableParam: string) => {
+    if (!(transferableParam in queryStrings)) {
+      queryStringsWithTransferedParams[transferableParam] = urlSearchParams[transferableParam]
+    }
+  })
+  return queryStringsWithTransferedParams
```
prilutskiy, line 31: Please update this code to use built-in `URL` type. Like this:

```javascript
const url = new URL(window.location.href)
// ...
url.searchParams.has('my_param')
// ...
url.searchParams.set('my_param', 'my_param_value')
```

packages/embed/src/utils/url-parameters-transfer.ts
```diff
@@ -0,0 +1,32 @@
+type Params = { [key: string]: string }
+
+type QueryStrings = {
+  [key: string]: any
+}
```
prilutskiy, line 5: Please put type declarations below the meaningful code

prilutskiy (changes requested):

packages/embed/src/utils/url-parameters-transfer.ts
```diff
@@ -0,0 +1,32 @@
...
+    var vars = search.split('&')
+    for (var i = 0; i < vars.length; i++) {
+      var pair = vars[i].split('=')
+      params[pair[0] as keyof Params] = decodeURIComponent(pair[1])
+    }
+  }
+  return params
+}
+
+export const transferUrlParametersToQueryStrings = (
```
mathio, line 19: URL has just one query string. So you are not actually generating "query strings" but rather "query parameters". Please rename the method, my suggestion would be `getTransferableQueryParams`.

Please name the file the same as its main method, eg. if the method is `getTransferableQueryParams()` name the file `get-transferable-query-params.ts`.

mathio, line 19: Is this still a work in progress PR? I dont see the method actually used anywhere. Please include an example in `demo-html` and a functional test (specify transferable params, visit the demo page with those params in query string and check it is also present in typeform iframe src).

mathio (changes requested):

packages/embed/src/utils/url-parameters-transfer.ts
mathio: If you use built-in `URL` type you might not need custom types at all.

packages/embed/src/utils/url-parameters-transfer.ts
antoniofull: The file and all the functions were actually named like this, I simply imported the code and added ts

antoniofull: Where should I add the example? React, HTML all of them?
Why functional tests? I think this unit test is enough. We just need to check the code, not the UI or the user actions.

mathio: I think HTML is sufficient. You are right - since there is not UI we dont need visual tests (but still should have functional test). Unit tests confirms the method works correctly. Functional test confirms the function is implemented correctly (you only need to check the iframe src is correct, no need to check how renderer handles it further).

mathio: We are trying to improve the lib, so we should rename things where appropriate.

PR #178, opened 2021-02-22
feat(DIST-699): Pass default value for source and medium

mathio:
Default source is domain (without the www) and medium is 'embed-sdk'.


packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -22,19 +34,14 @@ const mapOptionsToQueryParams = (type: EmbedType, options: UrlOptions): Record<s
...
-  const queryParams = mapOptionsToQueryParams(type, options)
-  Object.entries(queryParams).forEach(([paramName, paramValue]) => {
-    if (!paramValue) {
-      return
-    }
-    url.searchParams.set(paramName, paramValue)
-  })
+  const url = new URL(`https://form.typeform.com/to/${formId}`)
+  Object.entries(queryParams)
+    .filter(([_paramName, paramValue]) => paramValue !== undefined && paramValue !== null)
```
mathio, line 41: This is to avoid removing zero value which is still a valid value and we want to pass it to query params but it evaluates as falsy.

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -8,7 +19,8 @@ const typesToEmbed: Record<EmbedType, string> = {
   'side-tab': 'popup-side-panel',
 }
 
-const mapOptionsToQueryParams = (type: EmbedType, options: UrlOptions): Record<string, any> => ({
+const mapOptionsToQueryParams = (type: EmbedType, embedId: string, options: UrlOptions): Record<string, any> => ({
```
mathio, line 22: Lets keep all the logic of mapping options (or other values) to query params in one method.

packages/embed/src/utils/build-iframe-src.ts
prilutskiy: As a fan of short syntax, I must suggest the following:

```suggestion
    .filter(([, paramValue]) => paramValue != null)
```

packages/embed/src/utils/remove-undefined-keys.ts
```diff
@@ -0,0 +1,5 @@
+export const removeUndefinedKeys = (obj: Record<string, any>): Record<string, any> => {
+  return Object.entries(obj)
+    .filter(([_key, value]) => value !== undefined && value !== null)
```
prilutskiy, line 3:
```suggestion
    .filter(([, value]) => value != null)
```

packages/embed/src/utils/build-iframe-src.ts
mrisoli: You can do !! to check for a falsy value, also, why not just do a non-strict check for null which typescript should handle for both undefined and null?
Also, I'd consider removing the filter and doing the check in the forEach since the entries are not reused as that allocates extra memory
```typescript
Object.entries(queryParams).forEach([key, value] => {
  if (!!value) {
    url.searchParams.set(paramName, paramValue)
  }
})
```

mrisoli (approved):

packages/embed/src/utils/remove-undefined-keys.ts
mathio: I know `!= null`  catches only `null` and `undefined` but I really hate how non-strict comparison in javascript can lead to unexpected results. Not really in this case, but you need to have deeper understanding of the language. Shorthands like this also lead to "why not just `!value`?" which also catches empty string and zero - and thats obviously incorrect.

packages/embed/src/utils/build-iframe-src.ts
mathio: `!!value` prevents us from adding empty strings or zeros which is incorrect.

PR #179, opened 2021-02-23
feat(DIST-700): Add transitiveSearchParams

antoniofull:
feat(DIST-700): Add transferableUrls parameters

Adds transferableURL parameters
- [x] Link to the issue - if any [DIST-700](https://jira.typeform.tf/browse/DIST-700)

### DIFF
--- a/packages/demo-html/public/widget-html.html
+++ b/packages/demo-html/public/widget-html.html
@@ -12,7 +12,13 @@
     </style>
   </head>
   <body>
-    <div id="wrapper" data-tf-widget="moe6aa" data-tf-medium="unit-test"></div>
+    <div
+      id="wrapper"
+      data-tf-widget="moe6aa"
+      data-tf-medium="unit-test"
+      data-tf-transitive-search-params="foo,bar"
+      data-tf-source="localhost"
+    ></div>
     <script src="./lib/embed-next.js"></script>
   </body>
 </html>
--- a/packages/demo-html/public/widget-js.html
+++ b/packages/demo-html/public/widget-js.html
@@ -16,7 +16,12 @@
     <div id="wrapper"></div>
     <script src="./lib/embed-next.js"></script>
     <script>
-      window.tf.createWidget("moe6aa", { container: document.getElementById("wrapper"), medium: "unit-test" });
+      window.tf.createWidget("moe6aa", {
+        container: document.getElementById("wrapper"),
+        source: "localhost",
+        transitiveSearchParams: ["foo", "bar"],
+        medium: "unit-test",
+      });
     </script>
   </body>
 </html>
--- a/packages/embed/src/utils/build-iframe-src.ts
+++ b/packages/embed/src/utils/build-iframe-src.ts
@@ -2,6 +2,7 @@ import { EmbedType, UrlOptions } from '../base'
 
 import { removeUndefinedKeys } from './remove-undefined-keys'
 import { isDefined } from './is-defined'
+import { getTransitiveSearchParams } from './get-transitive-search-params'
 
 const defaultUrlOptions: UrlOptions = {
   source: window?.location?.hostname.replace(/^www\./, ''),
@@ -20,30 +21,33 @@ const typesToEmbed: Record<EmbedType, string> = {
   'side-tab': 'popup-side-panel',
 }
 
-const mapOptionsToQueryParams = (type: EmbedType, embedId: string, options: UrlOptions): Record<string, any> => ({
-  'typeform-embed-id': embedId,
-  'typeform-embed': typesToEmbed[type],
-  'typeform-source': options.source,
-  'typeform-medium': options.medium,
-  'typeform-medium-version': options.mediumVersion,
-  'embed-hide-footer': options.hideFooter ? 'true' : undefined,
-  'embed-hide-headers': options.hideHeaders ? 'true' : undefined,
-  'embed-opacity': options.opacity,
-  'disable-tracking': options.disableTracking ? 'true' : undefined,
-})
+const mapOptionsToQueryParams = (type: EmbedType, embedId: string, options: UrlOptions): Record<string, any> => {
+  const transitiveParams = getTransitiveSearchParams(options.transitiveSearchParams)
+  const params = {
+    'typeform-embed-id': embedId,
+    'typeform-embed': typesToEmbed[type],
+    'typeform-source': options.source,
+    'typeform-medium': options.medium,
+    'typeform-medium-version': options.mediumVersion,
+    'embed-hide-footer': options.hideFooter ? 'true' : undefined,
+    'embed-hide-headers': options.hideHeaders ? 'true' : undefined,
+    'embed-opacity': options.opacity,
+    'disable-tracking': options.disableTracking ? 'true' : undefined,
+  }
+  return { ...params, ...transitiveParams }
+}
 
 export const buildIframeSrc = (params: BuildIframeSrcOptions): string => {
   const { formId, type, embedId, options } = params
-
   const queryParams = mapOptionsToQueryParams(type, embedId, addDefaultUrlOptions(options))
 
   const url = new URL(`https://form.typeform.com/to/${formId}`)
+
   Object.entries(queryParams)
     .filter(([, paramValue]) => isDefined(paramValue))
     .forEach(([paramName, paramValue]) => {
       url.searchParams.set(paramName, paramValue)
     })
-
   return url.href
 }
 
--- a/packages/embed/src/utils/get-transitive-search-params.ts
+++ b/packages/embed/src/utils/get-transitive-search-params.ts
@@ -1,0 +1,14 @@
+export const getTransitiveSearchParams = (transitiveSearchParams?: string[]) => {
+  const url = new URL(window.location.href)
+  const queryParamsWithTransitiveParams = {}
+
+  if (transitiveSearchParams && transitiveSearchParams.length > 0) {
+    transitiveSearchParams.forEach((key: string) => {
+      if (url.searchParams.has(key)) {
+        queryParamsWithTransitiveParams[key] = url.searchParams.get(key)
+      }
+    })
+  }
+
+  return queryParamsWithTransitiveParams
+}
--- a/packages/embed/src/utils/load-options-from-attributes.spec.ts
+++ b/packages/embed/src/utils/load-options-from-attributes.spec.ts
@@ -34,7 +34,7 @@ describe('load-options-from-attributes', () => {
       })
     })
 
-    describe('to booolean', () => {
+    describe('to boolean', () => {
       it('should transform "yes" to true', () => {
         expect(transformAttributeValue('yes', 'boolean')).toBe(true)
       })
@@ -79,6 +79,24 @@ describe('load-options-from-attributes', () => {
       it('should transform empty string (attribute with no value) to undefined', () => {
         expect(transformAttributeValue('', 'integer')).toBe(undefined)
       })
+    })
+  })
+
+  describe('to array', () => {
+    it('should transform string as array', () => {
+      expect(transformAttributeValue('a, b', 'array')).toEqual(['a', 'b'])
+    })
+
+    it('should remove empty spaces around text', () => {
+      expect(transformAttributeValue('foo , bar , test', 'array')).toEqual(['foo', 'bar', 'test'])
+    })
+
+    it('should transform malformed string in empty array', () => {
+      expect(transformAttributeValue(',', 'array')).toEqual([])
+    })
+
+    it('should return undefined if no value', () => {
+      expect(transformAttributeValue('', 'array')).toEqual(undefined)
     })
   })
 
--- a/packages/embed/.prettierrc
+++ b/packages/embed/.prettierrc
@@ -2,5 +2,6 @@
   "trailingComma": "es5",
   "tabWidth": 2,
   "semi": false,
-  "singleQuote": true
+  "singleQuote": true,
+  "printWidth": 120
 }
--- a/packages/embed/e2e/spec/functional/widget.spec.ts
+++ b/packages/embed/e2e/spec/functional/widget.spec.ts
@@ -1,16 +1,42 @@
 describe('Widget', () => {
-  before(() => {
-    cy.visit('/widget-js.html')
-  })
-
-  it('should display widget', () => {
-    cy.get('.typeform-widget iframe').should('be.visible')
-    cy.get('.typeform-widget iframe').invoke('attr', 'src').should('contain', 'form.typeform.com/to/')
-  })
-
-  it('should pass options as query param', () => {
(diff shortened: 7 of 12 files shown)


packages/embed/e2e/spec/functional/widget-with-parameters.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  it('should display widget', () => {
+    cy.get('.typeform-widget iframe').should('be.visible')
+    cy.get('.typeform-widget iframe').invoke('attr', 'src').should('contain', 'form.typeform.com/to/')
+  })
+
+  it('should pass params from options to the iframe', () => {
+    cy.get('.typeform-widget iframe')
+      .invoke('attr', 'src')
+      .should('contain', 'typeform-embed=embed-widget&typeform-source=localhost&foo=foo')
```
mathio, line 14: Great and simple test 👍

packages/embed/e2e/spec/functional/widget-with-parameters.spec.ts
```diff
@@ -0,0 +1,16 @@
+describe('Widget with parameters', () => {
+  before(() => {
+    cy.visit('/widget-with-params-js.html?foo=foo')
```
mathio, line 3: I dont think we need this as a separate file, we can reuse existing `demo-js.html`. I dont want the collection of demos to grow to many files like it did in the old embed (and it was a mess).

packages/embed/src/utils/get-transferable-query-params.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  const location: Location = window.location
+
+  beforeAll(() => {
+    delete (window as any).location
+
+    window.location = {
+      search: '?foo=jason&bar=rachel&utm_medium=cpc&utm_campaign=camp2008&utm_source=instagram&embed-hide-footer=false',
+      href:
+        'http://localhost?foo=jason&bar=rachel&utm_medium=cpc&utm_campaign=camp2008&utm_source=instagram&embed-hide-footer=false',
+    } as any
```
mathio, line 13: I suggest to define the query string just once, I think it increases the readability since it is quite long:

```suggestion
    const search = ''?foo=jason&bar=rachel&utm_medium=cpc&utm_campaign=camp2008&utm_source=instagram&embed-hide-footer=false''

    window.location = {
      search,
      href: 'http://localhost`${search},
    } as any
```

packages/embed/src/utils/get-transferable-query-params.ts
```diff
@@ -0,0 +1,18 @@
+export const getTransferableQueryParams = (queryStrings: QueryStrings, transferableUrlParameters?: string[]) => {
```
mathio, line 1: This is not "query string", but rather "query params".
```suggestion
export const getTransferableQueryParams = (queryParams: Record<string, any>, transferableUrlParameters?: string[]) => {
```

prilutskiy (approved):

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -29,7 +31,8 @@ export const buildIframeSrc = (params: BuildIframeSrcOptions): string => {
   }
 
   const queryParams = mapOptionsToQueryParams(type, options)
-  Object.entries(queryParams).forEach(([paramName, paramValue]) => {
+  const queryWithTransferableParams = getTransferableQueryParams(queryParams, options.transferableUrlParameters)
```
mathio, line 34: I think this belongs to the `mapOptionsToQueryParams`, since we already pass `options.transferableUrlParameters` to it. I would like to keep all this query building logic in one place.

Also, we could simplify `getTransferableQueryParams` to not check if the param already exists in `queryParams`, we can just override those params:

```
const mapOptionsToQueryParams = (type: EmbedType, options: UrlOptions): Record<string, any> => {
  const transferredParams = getTransferableQueryParams(options.transferableUrlParameters)
  const params = {
    'typeform-embed': typesToEmbed[type],
    'typeform-source': options.source,
    // etc...
  }
  return { ...transferredParams, ...params }
}
```

packages/demo-html/public/widget-with-params-js.html
```diff
@@ -0,0 +1,26 @@
...
+    <link rel="stylesheet" href="./lib/css/widget.css" />
+  </head>
+  <body>
+    <div id="wrapper"></div>
+    <script src="./lib/embed-next.js"></script>
+    <script>
+      window.tf.createWidget("moe6aa", {
+        container: document.getElementById("wrapper"),
+        source: "localhost",
+        transferableUrlParameters: ["foo"],
```
mathio, line 22: Can we do an example of using pure HTML (in `demo-html.html`) too?

Since this is an array you would need to extend `Transformation` in `load-options-from-attributes.ts` to also include `'array'` and parse an array out of a coma separated list - add new `case` to `transformAttributeValue`. Then the demo would look like this:

```
<div ... data-tf-transferable-url-parameters="foo,bar"></div>
```

mathio (changes requested):
Before you merge this please update the commit message to `Add transferableUrlparameters`

mathio:
Also, I really hate the name "transferable URL parameters" (and I always hated it before). This name feels very formal, long and unnatural. Can we some up with a better and shorter name?

Maybe `readFromUrl`? `loadFromUrl`?

antoniofull:
> Also, I really hate the name "transferable URL parameters" (and I always hated it before). This name feels very formal, long and unnatural. Can we some up with a better and shorter name?
> 
> Maybe `readFromUrl`? `loadFromUrl`?

If you hate it  why did you let it add in the first place? This code isn't old at all xDd

antoniofull:
> Also, I really hate the name "transferable URL parameters" (and I always hated it before). This name feels very formal, long and unnatural. Can we some up with a better and shorter name?
> 
> Maybe `readFromUrl`? `loadFromUrl`?

`mergeUrlParams`?

prilutskiy:
@antoniofull @mathio how about `transientSearchParams` ?

mathio:
@antoniofull well, it was 7 months ago and now I know better :) And since we are working on a new version that does not require backward compatibility I would like to take this opportunity.

Since typeform URL params are kind of hidden on the host page and the only params visible to the user I like to think about this functionality it is actually reading (or loading) data from the host page URL parameters. Yes, the implementation is to "merge", but I'd rather name it using something that user sees - and they see their typeform will load data (eg. hidden fields) from host page query params. I am open to `transferParams` or maybe even `transferUrlParams` as a last option :)

mathio:
@prilutskiy _**transient**_
* _adjective: lasting only for a short time; impermanent._
* _noun: a person who is staying or working in a place for a short time only._
* _noun: a momentary variation in current, voltage, or frequency._

That does not really make sense to me 🤔 It does not last short time.

mathio:
`injectParams` because you instruct embed lib to inject those params from page URL into embedded typeform? Sorry just brainstoriming here in the PR 😃

prilutskiy:
@mathio sorry, I meant `transitive` :D

mathio:
@prilutskiy in that case `transitiveParams` to make it shorter? :)

prilutskiy:
@mathio `transitiveParams` sounds nice! `transitiveSearchParams` might also be a good option because it tells that it's about search params and URL, rather than just about function params or something.

mathio:
True. Lets go with `transitiveSearchParams` if there are no objections or other proposals.

packages/embed/src/utils/build-iframe-src.ts
antoniofull: yep, makes sense

packages/demo-html/public/widget-with-params-js.html
antoniofull: cool!

antoniofull:
@mathio @prilutskiy #everytransitiveSearchParamsCounts 🚀

prilutskiy (approved):

packages/demo-html/public/widget-html.html
```diff
@@ -12,7 +12,13 @@
     </style>
   </head>
   <body>
-    <div id="wrapper" data-tf-widget="moe6aa" data-tf-medium="unit-test"></div>
+    <div
+      id="wrapper"
+      data-tf-widget="moe6aa"
+      data-tf-medium="unit-test"
+      data-tf-transferable-url-parameters="foo,bar"
```
mathio, line 19: Should be `data-tf-transitive-search-params`

packages/demo-html/public/widget-js.html
```diff
@@ -16,7 +16,12 @@
     <div id="wrapper"></div>
     <script src="./lib/embed-next.js"></script>
     <script>
-      window.tf.createWidget("moe6aa", { container: document.getElementById("wrapper"), medium: "unit-test" });
+      window.tf.createWidget("moe6aa", {
+        container: document.getElementById("wrapper"),
+        source: "localhost",
+        transferableUrlParameters: ["foo"],
```
mathio, line 22: Should be `transitiveSearchParams`, please include 2 types here to demonstrate array types work (`"foo", "bar"` like in `widget-html.html`).

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -19,17 +20,21 @@ const typesToEmbed: Record<EmbedType, string> = {
...
-  'typeform-source': options.source,
-  'typeform-medium': options.medium,
-  'typeform-medium-version': options.mediumVersion,
-  'embed-hide-footer': options.hideFooter ? 'true' : undefined,
-  'embed-hide-headers': options.hideHeaders ? 'true' : undefined,
-  'embed-opacity': options.opacity,
-  'disable-tracking': options.disableTracking ? 'true' : undefined,
-})
+const mapOptionsToQueryParams = (type: EmbedType, embedId: string, options: UrlOptions): Record<string, any> => {
+  const transferredParams = getTransitiveSearchParams(options.transferableUrlParameters as string[])
```
mathio, line 24: `options.transferableUrlParameters` should be `options.transitiveSearchParams`

I suggest to also rename `transferredParams` to `transitiveParams`.

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -42,6 +47,17 @@ export const buildIframeSrc = (params: BuildIframeSrcOptions): string => {
     .forEach(([paramName, paramValue]) => {
       url.searchParams.set(paramName, paramValue)
     })
+  if (embedId) {
+    url.searchParams.set('typeform-embed-id', embedId)
```
mathio, line 51: This is already added in `mapOptionsToQueryParams`, maybe this was just a bad rebase on latest master.

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -42,6 +47,17 @@ export const buildIframeSrc = (params: BuildIframeSrcOptions): string => {
...
+  }
+
+  const queryWithTransitiveSearchParams = getTransitiveSearchParams(queryParams, options.transferableUrlParameters)
```
mathio, line 54: Do we need this? I think `mapOptionsToQueryParams` already takes care of adding transitive search params.

packages/embed/src/utils/get-transitive-search-params.ts
```diff
@@ -0,0 +1,14 @@
+export const getTransitiveSearchParams = (queryParams: Record<string, any>, transitiveSearchParams?: string[]) => {
+  const url = new URL(window.location.href)
+  const queryStringsWithSearchParams = { ...queryParams }
```
mathio, line 3: This is not "query strings", there is just one "query string" in a URL. Maybe `queryParamsWithTransitiveParams`?

packages/embed/src/utils/load-options-from-attributes.spec.ts
```diff
@@ -82,6 +82,12 @@ describe('load-options-from-attributes', () => {
     })
   })
 
+  describe('to array', () => {
```
mathio, line 85: Please also include tests for malformed strings, eg if it starts or ends with coma (`,`), if it contains white spaces around comas or inside field name.

mathio (changes requested):

packages/embed/e2e/spec/functional/widget-with-parameters.spec.ts
```diff
@@ -0,0 +1,16 @@
+describe('Widget with parameters', () => {
+  before(() => {
+    cy.visit('/widget-js.html?foo=foo')
```
mathio, line 3: Please add `bar=bar` too since it is on the list. And also 1 query param (eg.`baz=baz` that is not in the list and check its not in src

```
.invoke('attr', 'src')
.should('not.contain', 'baz')
```

Also please add the same test for `widget-html.html?...` to demonstrate it works too.

packages/demo-html/public/widget-html.html
```diff
@@ -12,7 +12,13 @@
     </style>
   </head>
   <body>
-    <div id="wrapper" data-tf-widget="moe6aa" data-tf-medium="unit-test"></div>
+    <div
+      id="wrapper"
+      data-tf-widget="moe6aa"
+      data-tf-medium="unit-test"
+      data-tf-transitive-url-parameters="foo,bar"
```
mathio, line 19: Typo? Should be `data-tf-transitive-search-parameters`?

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -20,17 +21,21 @@ const typesToEmbed: Record<EmbedType, string> = {
...
-  'typeform-source': options.source,
-  'typeform-medium': options.medium,
-  'typeform-medium-version': options.mediumVersion,
-  'embed-hide-footer': options.hideFooter ? 'true' : undefined,
-  'embed-hide-headers': options.hideHeaders ? 'true' : undefined,
-  'embed-opacity': options.opacity,
-  'disable-tracking': options.disableTracking ? 'true' : undefined,
-})
+const mapOptionsToQueryParams = (type: EmbedType, embedId: string, options: UrlOptions): Record<string, any> => {
+  const transitiveParams = getTransitiveSearchParams(options.transitiveSearchParams as string[])
```
mathio, line 25: Is the cast `as string[]` necessary? I see it is defined correctly as `transitiveSearchParams?: string[]` and the method accepts same type as parameter.

packages/demo-html/public/widget-html.html
antoniofull: yeah might be, maybe when I rebased

packages/embed/e2e/spec/functional/widget-with-parameters.spec.ts
antoniofull: maybe we can get rid of `widget-with-parameters.spec` and just keep all tests for widget in one single file? WDYT

antoniofull:
@mathio updated with comments review

mathio (approved):

PR #180, opened 2021-02-23
feat: Add helper method isDefined

mathio:
This is a [custom type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), I recommend to read linked docs if you are not familiar with the concept.

We should use this check in place of `value != null` or `!!value` as those approaches might lead to unexpected behavior (eg. the latter will return false for `0` which is in most cases valid value).


prilutskiy (approved):

mathio:
I thought this was already merged :)

PR #182, opened 2021-02-24
feat(DIST-725): Add popover

mrisoli:
## Description

Add popover factory
Move create embed logic to generic factory for all embeds
Update tests and properties


mathio (changes requested):
I dont like the unified `createEmbed` approach. I am afraid is too early to refactor it this way and we will end up with too many logic branches in the code.

"create" methods share a lot of the same code now, but that will change as we add styles and other functionality.

"initialize" methods share a lot of code too and we might end up refactoring this, but only after all embed types are implemented.

mrisoli:
closed, new one is #195

PR #183, opened 2021-02-24
feat(DIST-712): Add popup loader and styles

mathio:


packages/embed/src/factories/create-popup/create-popup.spec.ts
```diff
@@ -1,32 +1,37 @@
 import { createPopup, Popup } from './create-popup'
 
+jest.useFakeTimers()
```
prilutskiy, line 3: what's this?

prilutskiy (approved):

packages/embed/src/factories/create-popup/create-popup.spec.ts
mathio: https://jestjs.io/docs/en/timer-mocks

When you use `setTimeout` in your code, it will get executed when you call `jest.runAllTImers()`. You dont have to actually wait.

mrisoli (approved):

PR #185, opened 2021-02-25
feat: Add Next.js demo (with server side rendering)

mathio:


packages/demo-nextjs/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+```javascript
+import { createWidget } from "@typeform/embed";
+import "@typeform/embed/build/css/widget.css"; // import necessary CSS
+
+const MyTypeformEmbed = () => {
+  const container = useRef();
+
+  useEffect(() => {
+    createWidget("<form id>", { container: container.current });
```
prilutskiy, line 17: I think I saw somewhere that container was renamed to element. Was it? Not sure though.

mathio: That PR was not merged. It is container.

prilutskiy: Ok, then we're cool.

prilutskiy (approved):

mrisoli (approved):

PR #187, opened 2021-02-25
feat: Deploy to AWS S3

mathio:
🤞

### DIFF
--- a/packages/embed/webpack.config.js
+++ b/packages/embed/webpack.config.js
@@ -1,10 +1,14 @@
 const path = require('path')
 
 const sass = require('node-sass')
+const webpack = require('webpack')
 const CopyPlugin = require('copy-webpack-plugin')
 
+const mode = process.env.NODE_ENV === 'production' ? 'production' : 'development'
+const isProd = mode === 'production'
+
 const baseConfig = {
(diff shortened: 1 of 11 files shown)


mathio:
Sorry for the spam. AWS test deploy successful: https://embed.typeform.com/next/embed.js (only 17kB, compared to 164 kB of old embed lib)

Example: https://glitch.com/edit/#!/embed-next-html (the CSS is not injected correctly at the moment)

packages/embed/webpack.config.js
```diff
@@ -20,20 +24,26 @@ const baseConfig = {
...
+      },
+    }),
   ],
 }
 
 const npmConfig = {
   ...baseConfig,
   entry: './src/index.ts',
-  devtool: 'inline-source-map',
+  devtool: isProd ? 'inline-source-map' : false,
```
prilutskiy, line 46: Are you sure about this setting? It might be helpful for debugging purposes.

mathio: You are right, it should be the other way. Keep source maps for development, remove it only in production. It is a question of 13kB vs 50kB in lib size :)

PR #189, opened 2021-02-26
feat: Add size, width and height to options

mathio:
You can specify popup and slider size now.

### DIFF
--- a/packages/embed/src/factories/create-popup/create-popup.ts
+++ b/packages/embed/src/factories/create-popup/create-popup.ts
@@ -1,4 +1,5 @@
-import { createIframe, hasDom } from '../../utils'
+import { createIframe, hasDom, isDefined } from '../../utils'
+import { POPUP_SIZE } from '../../constants'
 
 import { PopupOptions } from './popup-options'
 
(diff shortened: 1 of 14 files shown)


packages/embed/src/factories/create-slider/create-slider.ts
```diff
@@ -58,12 +54,12 @@ export const createSlider = (formId: string, userOptions: SliderOptions): Slider
     }
   }
 
-  const { position = 'right', ...options } = { ...defaultSliderOptions, ...userOptions }
+  const { position = 'right', width = 800, ...options } = userOptions
```
prilutskiy, line 57: let's extract this into a constant and place in `constants/`

packages/embed/src/factories/create-popup/create-popup.ts
```diff
@@ -58,12 +61,12 @@ export const createPopup = (formId: string, userOptions: PopupOptions): Popup =>
     }
   }
 
-  const options = { ...defaultPopupOptions, ...userOptions }
+  const { width, height, size = 100, ...options } = userOptions
```
prilutskiy, line 64: let's extract this into a constant and place in `constants/`

prilutskiy (approved):

PR #193, opened 2021-03-02
chore: Testing library dom

scarciofolomarco:


prilutskiy (approved):

mathio:
@scarciofolomarco  I dont have any experience with this lib. What are the benefits?

scarciofolomarco:
> @scarciofolomarco I dont have any experience with this lib. What are the benefits?

If you have used react testing library (which I think all do), its the same API but rendering dom elements instead of react components

PR #195, opened 2021-03-03
feat(DIST-725): Add popover

mrisoli:
## Description

Adds popover widget to next branch


antoniofull (approved):

packages/demo-html/public/popover-html.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Static HTML Demo</title>
+    <style>
+      #wrapper {
```
mathio, line 7: `#wrapper` is not required for this page nor for `popover-js.html`.

packages/embed/src/factories/create-popover/popover-options.ts
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @type {string}
+     */
+    buttonColor?: string
+    /**
+     * Specify the image url for the popover trigger button icon
+     *
+     * @type {string}
+     */
+    customIcon?: string
```
mathio, line 47: We should also load those options from attributes in [loadOptionsFromAttributes](https://github.com/Typeform/embed/blob/6d7c673b0694318525c5b9774f885c319f29eb10/packages/embed/src/utils/load-options-from-attributes.ts#L58).

mrisoli: Should they be included if they are popover specific?

mrisoli: nvm now I understood how those options are loaded

PR #205, opened 2021-03-23
fix(DIST-682): Send post message only if the form is present on page

mathio:
We need to figure out how to release this as `0.*` version, not `1.*`.

### DIFF
--- a/demo/widget-tabs.html
+++ b/demo/widget-tabs.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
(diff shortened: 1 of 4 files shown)


prilutskiy (approved):

prilutskiy:
Please add a unit test :)

antoniofull (approved):

mathio:
@prilutskiy its not tested now and its hard to mock. I'd rather not given we already have new embed version out :) You can test it out manually in `demo/widget-tabs.html `.

PR #212, opened 2021-04-08
Fix onSubmit type

yuya-fujimoto:
Adding `payload: { responseId: string }` in onSubmit type

### DIFF
--- a/packages/embed/src/base/actionable-options.ts
+++ b/packages/embed/src/base/actionable-options.ts
@@ -7,7 +7,7 @@ export type ActionableOptions = {
   /**
    * Callback function that will be executed right after the typeform is successfully submitted.
    */
-  onSubmit?: () => void
+  onSubmit?: (payload: { responseId: string }) => void
   /**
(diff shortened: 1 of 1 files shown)


mathio:
Thank you for the PR. 

Failing check `Secrets scan` is irrelevant, its just missing authorization for the service that runs this check.

However I can see your commit message does not pass our commitlint requirements. A check should be run in commit hook (before commit) unless you skip it:

```
husky > commit-msg (node v12.21.0)
⧗   input: fix onSubmit type
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
⚠   scope may not be empty [scope-empty]

✖   found 2 problems, 1 warnings
```

yuya-fujimoto:
My bad for missing commitlint. Updated the commit message.

mathio (approved):

mathio:
@yuya-fujimoto thank you! And also thank you for opening 1st PR from outside of the company, I was able to update our Github Actions to skip those that dont work with PRs from repo forks.

PR #218, opened 2021-04-15
feat(DIST-866): deploy next version in non-preview mode

prilutskiy:


mathio (approved):

mathio:
How can we check if the deployment was successful?

prilutskiy:
> How can we check if the deployment was successful?

Jarvis should throw if it's not; also the deployment event is sent at the end of the queue, so it won't be sent in case of failure.

PR #222, opened 2021-04-15
fix(DIST-858): Popup position

mathio:
We have to use flexbox instead of `translate(-50%, -50%)` to avoid blur.


mathio:
This fix broke other embed types 🤦‍♂️

mrisoli (approved):

PR #225, opened 2021-04-22
fix(DIST-864): Calculate mobile modal height for popup

mrisoli:
## Description

Use a ref to calculate a height based on the offsetTop of mobile modal
When its displaced by an absolute positioned header
Since the absolute header was causing the submit button sticky footer to be displayed outside the viewport

## Links

[DIST-864](https://jira.typeform.tf/browse/DIST-864)

### DIFF
--- a/src/core/views/mobile-modal.js
+++ b/src/core/views/mobile-modal.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { Component, createRef } from 'react'
 import PropTypes from 'prop-types'
 import styled, { createGlobalStyle } from 'styled-components'
 
@@ -18,6 +18,7 @@ import {
 import { DEFAULT_AUTOCLOSE_TIMEOUT } from './popup'
 
 const Wrapper = styled.div`
+  height: ${({ height }) => (height ? height : '100%')};
   visibility: ${(p) => (p.open ? 'visible' : 'hidden')};
   opacity: ${(p) => (p.open ? 1 : 0)};
   background-color: ${(p) => p.backgroundColor};
@@ -25,10 +26,9 @@ const Wrapper = styled.div`
   z-index: 10001;
   left: 0 !important;
   right: 0 !important;
-  top: 0 !important;
+  top: 80px !important;
   bottom: 0 !important;
   overflow: hidden !important;
-  height: 100%;
   transition: all 400ms ease ${(props) => props.openDelay}s;
 `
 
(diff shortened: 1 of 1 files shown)


src/core/views/mobile-modal.js
```diff
@@ -18,17 +18,17 @@ import {
...
+  height: ${({ height }) => (height ? height : '100%')};
   visibility: ${(p) => (p.open ? 'visible' : 'hidden')};
   opacity: ${(p) => (p.open ? 1 : 0)};
   background-color: ${(p) => p.backgroundColor};
   position: fixed !important;
   z-index: 10001;
   left: 0 !important;
   right: 0 !important;
-  top: 0 !important;
+  top: 80px !important;
```
mathio, line 29: Why change of `top` when the tab bar seems to be at the bottom?

mrisoli: Ooops, this was done to emulate the same situation from the issue, undone

scarciofolomarco (approved):

src/core/views/mobile-modal.js
```diff
@@ -60,12 +61,19 @@ class MobileModal extends Component {
...
+    this.wrapperRef = createRef()
   }
 
   componentDidMount() {
     const originalViewportContent = setMobileMetaViewport()
     this.setState({ originalViewportContent })
 
+    if (this.wrapperRef.current) {
+      this.setState({
+        height: `${document.documentElement.clientHeight - this.wrapperRef.current.offsetTop}px`,
```
mathio, line 73: Same question here. You are using `offsetTop`, but the tab bar in screenshot is on the bottom. How does this work?

antoniofull: The customer situation is that they override our css for the wrapper adding a top of 80px, that created a good edge case for testing this. The wrapper offsetTop is the distance of the wrapper from the top of the viewport, in this case would be 80 because it is forced by the customer, in our ideal case would always be 0 because our popup if fixed and has a top:0; So this applies only when there is a case similar to that one, in other situation will always return 0

mathio: If customer did some modifications on their side I would suggest them to fix it themselves.

antoniofull: I agree, as discussed since the customer is doing custom css, we suggested a simple css fix for their edge case https://jira.typeform.tf/browse/DIST-864 see comment

mathio (changes requested):
Just blocking the auto-merge. Good job figuring this out 👍

PR #235, opened 2021-05-06
feat(DIST-828): Widget, popover, sidetab sizes params

scarciofolomarco:


packages/embed/src/factories/create-widget/widget-options.ts
```diff
@@ -3,5 +3,22 @@ import { ActionableOptions, BaseOptions, UrlOptions } from '../../base'
 export type WidgetOptions = BaseOptions &
   UrlOptions &
   ActionableOptions & {
+    /**
+     * Element to place the widget into.
+     *
+     * @type {HTMLElement}
+     */
     container: HTMLElement
+    /**
```
prilutskiy, line 12: Maybe these should be extracted into a type, like `SizeableOptions` ?

prilutskiy (approved):
Good job, clean and has tests 👍

packages/embed/src/factories/create-widget/widget-options.ts
scarciofolomarco: hmmm good suggestion, on it

mrisoli (approved):

packages/embed/src/factories/create-popover/create-popover.spec.ts
```diff
@@ -16,12 +15,14 @@ afterEach(() => {
 describe('#createSidetab', () => {
   describe('#open', () => {
     it('should open', () => {
+      popover = createPopover('formId')
```
mathio, line 18: You can use `beforeEach` to recreate this before each test.

Same for sidetab tests.

packages/embed/src/utils/set-element-size.ts
```diff
@@ -0,0 +1,11 @@
+export const setElementSize = (element: HTMLElement, { width, height }: { width?: number; height?: number }) => {
```
mathio, line 1: Please do not define interfaces inline like this, it is harder to read.

```suggestion
interface ElementSize {
  width?: number
  height?: number
}

export const setElementSize = (element: HTMLElement, { width, height }: ElementSize) => {
```

Or you can pass each one as separate variable.

packages/embed/src/utils/set-element-size.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+      const element = document.createElement('div')
+      expect(setElementSize(element, { width: 100, height: 100 })).toHaveStyle({
+        width: '100px',
+        height: '100px',
+      })
+    })
+
+    it('should return same element if no params', () => {
+      const element = document.createElement('div')
+      const newElement = setElementSize(element, { width: undefined, height: undefined })
```
mathio, line 27: Why `undefined`? I would specify values here to confirm the element does not change even if its resized.

packages/embed/src/initializers/initialize-popovers.ts
```diff
@@ -21,6 +21,8 @@ export const initializePopovers = () => {
       buildOptionsFromAttributes(button, {
         buttonColor: 'string',
         customIcon: 'string',
+        width: 'integer',
```
mathio, line 24: I think we should get rid of `additionalAttributes` from `buildOptionsFromAttributes` and just build all attributes for all embed types. I think there is no harm building unused options and it will make the code easier to read. WDYT?

packages/embed/src/factories/create-widget/elements/build-widget.ts
```diff
@@ -1,6 +1,13 @@
-export const buildWidget = (iframe: HTMLIFrameElement): HTMLDivElement => {
+import { setElementSize } from '../../../utils'
+import { WidgetOptions } from '../widget-options'
+
+export const buildWidget = (iframe: HTMLIFrameElement, options: Omit<WidgetOptions, 'container'> = {}): HTMLElement => {
```
mathio, line 4: It is not necessary to send whole `Omit<WidgetOptions, 'container'>` here just to get width and height. We can pass each one as separate argument (as you did for `buildSidetab`).

packages/embed/src/utils/set-element-size.ts
mathio: Since we have `setElementSize` function extracted, lets use it in all places where we set element sizes -> popover and slider too.

mathio (changes requested):

packages/embed/src/utils/set-element-size.ts
scarciofolomarco: already did, thing is slider has a slightly different implementation also popover

packages/embed/src/factories/create-popover/create-popover.spec.ts
scarciofolomarco: it was like that, and I changed it because sometimes I need to pass different `options` depending on the test, so I had to `unmount` and `remount` the embed with the new options for those tests, which I think is not correct

packages/embed/src/utils/set-element-size.spec.ts
scarciofolomarco: I don't wanna make the 2nd argument optional, just the properties of the object, this test is to check it returns the same element when `width` nor `height` is`undefined`

packages/embed/src/utils/set-element-size.ts
mathio: Slider is doing this:

```
wrapper.style.width = `${width}px`
return wrapper
```

You could do 

```
return setElementSize(wrapper, { width })
```

mathio: Sorry I meant popup not popover. Something like this:

```
   if (isDefined(width) && isDefined(height)) {
-    wrapper.style.width = `${width}px`
-    wrapper.style.height = `${height}px`
+    return setElementSize(wrapper, {width, height)
   }
```

mathio:
Also important: dont forget to format commit message when squashing an merging :)

scarciofolomarco:
> Also important: dont forget to format commit message when squashing an merging :)

yeah, I usually amend but the merge commit won't let me

packages/embed/src/initializers/initialize-popovers.ts
scarciofolomarco: done!

packages/embed/src/factories/create-popover/create-popover.spec.ts
mathio: There already is `afterEach` that does `popover.unmount()`. If you dont want to do unmount in specific test you can restructure describes and have the `beforeEach` only for those where you need it.

```
describe('default popover', () => {
  beforeEach(() => {
    popover = createPopover('formId')
  })
  it('xxx', () => {
    popover.open()
    // ...
  })
})

describe('different popover', () => {
  it('xxx', () => {
    createPopover('formId', { buttonColor: white })
    // ...
  })
})
```

mathio:
> yeah, I usually amend but the merge commit won't let me

I usually amend too. To be able to amend I avoid merge commits and `git rebase master` instead.

packages/embed/src/factories/create-popover/create-popover.spec.ts
scarciofolomarco: hmmm great idea, on it

packages/embed/src/initializers/initialize-widgets.ts
```diff
@@ -16,6 +16,9 @@ export const initializeWidgets = () => {
     if (!formId) {
       throw new Error(`Invalid ${WIDGET_ATTRIBUTE}=${formId} for widget embed #${index}`)
     }
-    createWidget(formId, { ...buildOptionsFromAttributes(container), container })
```
mathio, line 19: Unnecessary change :) Its weird it passes linter and prettier in both cases.

packages/embed/src/factories/create-popup/create-popup.ts
```diff
@@ -36,8 +36,7 @@ const buildWrapper = (width?: number, height?: number, size?: number) => {
   wrapper.style.opacity = '0'
 
   if (isDefined(width) && isDefined(height)) {
-    wrapper.style.width = `${width}px`
-    wrapper.style.height = `${height}px`
+    return setElementSize(wrapper, { width, height })
```
mathio, line 39: I think returns here should be 

```
if (...) {
  return ...
} else {
  return ...
}
```

Or you can omit the else branch:

```
if (...) {
  return ...
}

return ...
```

mathio (approved):

PR #237, opened 2021-05-06
feat(DIST-945): Allow sandbox mode for embedded forms

### LINKED ISSUE #236: How to enable sandbox mode on an embedded form?
### Is  `__dangerous-disable-submissions` supported for embedded forms and, if it is, how can it be enabled it for an embedded form using this library in react?

I'm using the `createWidget` function in a react app to embed a form as per the [react demo](https://github.com/Typeform/embed/tree/main/packages/demo-react) and attempting to set `__dangerous-disable-submissions` to prevent form submission during development and UI testing. I believe I'm correctly setting the param according to the  [documentation](https://help.typeform.com/hc/en-us/articles/360029295952-Test-your-typeforms-without-creating-an-entry-sandbox-).

I've attempted to pass it via the `hidden` prop, but that resulted in the param being set as `[my other hidden fields]&__dangerous-disable-submissions=`. I assume the `=` at the end will not work.

I tried setting the prop manually by appending it to the iframe `src` attribute and calling `refresh` on the widget, but submissions are still being counted against my metrics (although the prop looks correct in the src uri).

mathio:
Fixes #236


antoniofull (approved):

packages/embed/src/utils/build-iframe.src.spec.ts
```diff
@@ -68,6 +69,7 @@ describe('build-iframe-src', () => {
           '&embed-opacity=50' +
           '&disable-tracking=true' +
           '&disable-auto-focus=true' +
+          '&__dangerous-disable-submissions=true' +
```
prilutskiy, line 72: a-ha! so this one is still useful ;)

prilutskiy (approved):

prilutskiy:
amazing feature 👏

packages/embed/src/utils/build-iframe.src.spec.ts
mathio: Yes and its also something we mention in our public docs, as I recently found out :)

PR #242, opened 2021-05-07
feat(DIST-779): React lib

mathio:


antoniofull (approved):

packages/embed-react/webpack.config.js
```diff
@@ -0,0 +1,46 @@
...
+      //   use: [
+      //     {
+      //       loader: require('styled-jsx/webpack').loader,
+      //       options: {
+      //         type: 'scoped',
+      //       },
+      //     },
+      //   ],
+      //   sideEffects: true,
+      // },
```
antoniofull, line 30: Is this needed?

mathio: Good catch, I think it is not.

scarciofolomarco (approved):

packages/embed-react/src/components/make-button-component.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const ref: MutableRefObject<GenericEmbed | null> = useRef(null)
+    useEffect(() => {
+      ref.current = createFn(id, props)
+      return () => ref.current?.unmount()
+    }, [id, props])
+
+    return (
+      <>
+        <InlineStyle filename={cssFilename} />
+        <button onClick={() => ref.current?.open()} className={className} style={style}>
```
mrisoli, line 34: Will the arrow function here not cause a rerender on every prop change?

packages/embed-react/src/components/make-button-component.tsx
```diff
@@ -0,0 +1,42 @@
...
+type CreateFn<T> = (id: string, props: CreateFnProps<T>) => GenericEmbed
+
+type GenericEmbed = {
+  unmount: () => void
+  open: () => void
+}
+
+function makeButtonComponent<T>(createFn: CreateFn<T>, cssFilename: string) {
+  return ({ id, children, style = {}, className = '', ...props }: ButtonComponentProps<T>) => {
+    const ref: MutableRefObject<GenericEmbed | null> = useRef(null)
```
mrisoli, line 25: I'm not a fan of the union with null here, I'd consider maybe having a default object instead, not saying necessarily that it is better, probably a matter of taste

packages/embed/src/factories/create-popup/create-popup.ts
```diff
@@ -97,7 +98,7 @@ export const createPopup = (formId: string, userOptions: PopupOptions = {}): Pop
       popup.style.opacity = '0'
       wrapper.style.opacity = '0'
       setTimeout(() => {
-        popup.parentNode.removeChild(popup)
+        popup.parentNode && popup.parentNode.removeChild(popup)
```
mrisoli, line 101: Could be replaced with a safe navigation operator here:

```suggestion
        popup.parentNode?.removeChild(popup)
```

scarciofolomarco: theres the `unmountElement` function too

packages/embed-react/src/components/make-button-component.tsx
mathio: Something like this?

```
const emptyEmbed: GenericEmbed = {
  unmount: () => {},
  open: () => {},
}
...
const ref: MutableRefObject<GenericEmbed> = useRef(emptyEmbed)
```

rix1:
Looking very much forward to having this merged! 🤩 Let me know if you need beta users 😁

prilutskiy (approved):

mathio:
React lib is released, please follow [the README](https://github.com/Typeform/embed/tree/main/packages/embed-react) for installation and usage. If there is anything to report, [please open a Github issue](https://github.com/Typeform/embed/issues). I will appreciate any testing.

cc: @rix1

PR #248, opened 2021-05-26
feat(DIST-923): Add notification dot to chat

antoniofull:
[Ticket DIST-923](https://jira.typeform.tf/browse/DIST-923)

Adds a Notification Dot next to the chat UI button
Screenshot 2021-05-26 at 12 31 48

### DIFF
--- a/packages/demo-html/public/chat.html
+++ b/packages/demo-html/public/chat.html
@@ -29,6 +29,7 @@
         data-tf-hidden="foo=foo value,bar=bar value"
         data-tf-opacity="50"
         data-tf-chat
+        data-tf-notification-days="3"
       ></div>
     </div>
 
@@ -45,6 +46,7 @@
       data-tf-popover="moe6aa"
       data-tf-medium="demo-test"
       data-tf-chat
+      data-tf-notification-days="3"
       data-tf-tooltip="Reallylonglongwidetooltipwordfortestingthebreakwordbehaviour"
     ></div>
 
--- a/packages/demo-html/public/popover-html.html
+++ b/packages/demo-html/public/popover-html.html
@@ -12,6 +12,7 @@
       data-tf-hidden="foo=foo value,bar=bar value"
       data-tf-tooltip="Live long and prosper 🖖 But what if the message here is too long and it wraps? Well, who are we to judge. Embed lib is lib-eral and lets you do whatever you want. Its the share page that imposes silly rules such as maximum tooltip message length."
       id="button"
+      data-tf-notification-days="3"
     ></div>
     <script src="./lib/embed-next.js"></script>
   </body>
--- a/packages/embed/e2e/spec/functional/chat.spec.ts
+++ b/packages/embed/e2e/spec/functional/chat.spec.ts
@@ -1,0 +1,27 @@
+describe(`Chat Popover`, () => {
+  before(() => {
+    cy.visit('/chat.html')
+  })
+
+  it('should display popover button on page load', () => {
+    cy.get('.typeform-popover-button').should('be.visible')
+  })
+
+  it('should display unread notification on the button', () => {
+    cy.get('[data-testid="typeform-popover-unread-dot"]').should('exist')
+  })
+
+  it('should open popover', () => {
+    cy.get('.typeform-popover-button').click()
+    cy.get('.typeform-popover-wrapper').should('be.visible')
+    cy.get('.typeform-popover-wrapper iframe').invoke('attr', 'src').should('contain', 'form.typeform.com/c/')
+    cy.get('[data-testid="typeform-popover-unread-dot"]').should('not.exist')
+  })
+
+  it('should close popover', () => {
+    cy.get('.typeform-popover-button').click()
+    cy.get('.typeform-popover-button [data-testid="default-icon"]').should('exist')
+    cy.get('.typeform-popover-button [data-testid="close-icon"]').should('not.exist')
+    cy.get('.typeform-popover-wrapper').should('not.exist')
+  })
+})
--- a/packages/embed/src/factories/create-popover/create-popover.spec.ts
+++ b/packages/embed/src/factories/create-popover/create-popover.spec.ts
@@ -4,8 +4,18 @@ import { createPopover, Popover } from './create-popover'
 
 let popover: Popover
 
+const mockedLocalStorage = (function () {
+  return {
+    getItem: jest.fn(),
+    setItem: jest.fn(),
+  }
+})()
+
 beforeEach(() => {
   jest.useFakeTimers()
+  Object.defineProperty(window, 'localStorage', {
+    value: mockedLocalStorage,
+  })
 })
 
 afterEach(() => {
@@ -122,6 +132,26 @@ describe('#createSidetab', () => {
         popover.open()
         jest.runAllTimers()
         expect(screen.queryByTestId('typeform-popover-tooltip')).toBeNull()
+      })
+    })
+
+    describe('#notificationDot', () => {
+      it('should render notificationDot', () => {
+        popover = createPopover('formId', { notificationDays: 1 })
+        expect(screen.getByTestId('typeform-popover-unread-dot')).toBeInTheDocument()
+      })
+
+      it('should hide notificationDot on form open', () => {
+        popover = createPopover('formId', { notificationDays: 2 })
+        fireEvent.click(screen.getByTestId('typeform-popover-button'))
+        jest.runAllTimers()
+        expect(screen.queryByTestId('typeform-popover-unread-dot')).toBeNull()
+      })
+
+      it('should store #notificationDot data in localStorage', () => {
+        popover = createPopover('formId', { notificationDays: 2 })
+        jest.runAllTimers()
+        expect(window.localStorage.setItem).toHaveBeenCalled()
       })
     })
   })
--- a/packages/embed/src/factories/create-popover/create-popover.ts
+++ b/packages/embed/src/factories/create-popover/create-popover.ts
@@ -97,6 +97,47 @@ const buildTooltip = (message: string, hide: () => void) => {
   return container
 }
 
+const LOCAL_STORAGE_NOTIFICATION_DATA = 'tfNotificationData'
+
+type LocalStorageCUIData = {
+  formId: string
+  expireDate: number
+}
+const today = new Date().getTime()
+
+const getLocalStorageData = () => {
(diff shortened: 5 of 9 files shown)


mrisoli (approved):

prilutskiy (approved):

packages/demo-html/public/popover-html.html
```diff
@@ -10,6 +10,7 @@
       data-tf-popover="moe6aa"
       data-tf-medium="demo-test"
       data-tf-hidden="foo=foo value,bar=bar value"
+      data-tf-notificationDot
```
prilutskiy, line 13:
```suggestion
      data-tf-notification-dot
```

prilutskiy (approved):

packages/demo-html/public/popover-html.html
antoniofull: hu good catch!

antoniofull: fixed 🙇

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -97,6 +97,48 @@ const buildTooltip = (message: string, hide: () => void) => {
...
+    // save data to localstorage
+    const expireDate = new Date()
+    expireDate.setDate(expireDate.getDate() + days)
+
+    saveToLocalstorage({
+      formId,
+      expireDate: expireDate.getTime(),
+    })
+  }
+  return true
```
mathio, line 131: Move this into `else` block please. Or just get rid of the `else` block completely.

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -97,6 +97,48 @@ const buildTooltip = (message: string, hide: () => void) => {
...
+  if (!data) return null
+  typeof window != 'undefined' && localStorage.setItem(LOCAL_STORAGE_NOTIFICATION_DATA, JSON.stringify(data))
+}
+
+function canBuildCUIDot(formId: string, days: number) {
+  const localStorageData = getLocalStorageData()
+  if (localStorageData && localStorageData.expireDate) {
+    return today > localStorageData.expireDate
+  } else {
+    // save data to localstorage
```
mathio, line 122: I would remove the comment. The method below says `saveToLocalstorage` too.

packages/demo-html/public/chat.html
```diff
@@ -45,6 +45,7 @@ <h1>this is an experimental feature</h1>
       data-tf-popover="moe6aa"
       data-tf-medium="demo-test"
       data-tf-chat
+      data-tf-notification-dot="3"
```
mathio, line 48: At first I thought this would be a number inside the dot. But its number of days to display the dot for, right? Maybe we could find a better name? `data-tf-notification-days`?

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -97,6 +97,48 @@ const buildTooltip = (message: string, hide: () => void) => {
...
+
+type LocalStorageCUIData = {
+  formId: string
+  expireDate: number
+}
+const today = new Date().getTime()
+
+const cuiLocaslStorageDate = typeof window !== 'undefined' && localStorage.getItem(LOCAL_STORAGE_NOTIFICATION_DATA)
+
+const getLocalStorageData = () => cuiLocaslStorageDate && JSON.parse(cuiLocaslStorageDate)
```
mathio, line 110: I would not access the variable from outer scope in a function, but rather read the value inside the function. I would also name it "data" not "date" (it is an object with form id and date).
```suggestion
const getLocalStorageData = () => {
  const data = typeof window !== 'undefined' && localStorage.getItem(LOCAL_STORAGE_NOTIFICATION_DATA)
  return data && JSON.parse(data)
}
```

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -97,6 +97,48 @@ const buildTooltip = (message: string, hide: () => void) => {
...
+
+function saveToLocalstorage(data: LocalStorageCUIData) {
+  if (!data) return null
+  typeof window != 'undefined' && localStorage.setItem(LOCAL_STORAGE_NOTIFICATION_DATA, JSON.stringify(data))
+}
```
mathio, line 115: Please capitalize the word `storage`:
```suggestion
function saveToLocalStorage(data: LocalStorageCUIData) {
  if (data && typeof window != 'undefined' ) {
    localStorage.setItem(LOCAL_STORAGE_NOTIFICATION_DATA, JSON.stringify(data))
  }
}
```

mathio: Also - is the `typeof window != 'undefined'` check necessary? Is the code ever executed server-side? If this is just for test we should mock it.

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -133,11 +177,30 @@ export const createPopover = (formId: string, userOptions: PopoverOptions = {}):
...
+      }, 250)
+    }
+  }
+
   if (options.tooltip && options.tooltip.length > 0) {
     tooltip = buildTooltip(options.tooltip, hideTooltip)
     popover.append(tooltip)
   }
 
+  if (options.chat && options.notificationDot) {
```
mathio, line 194: Can we enable this for all popovers? We will not expose it in share page, but get it working "for free" for all embeds when people integrate via SDK which I think is nice. Tooltip works form all embeds too.

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -97,6 +97,48 @@ const buildTooltip = (message: string, hide: () => void) => {
...
+const cuiLocaslStorageDate = typeof window !== 'undefined' && localStorage.getItem(LOCAL_STORAGE_NOTIFICATION_DATA)
+
+const getLocalStorageData = () => cuiLocaslStorageDate && JSON.parse(cuiLocaslStorageDate)
+
+function saveToLocalstorage(data: LocalStorageCUIData) {
+  if (!data) return null
+  typeof window != 'undefined' && localStorage.setItem(LOCAL_STORAGE_NOTIFICATION_DATA, JSON.stringify(data))
+}
+
+function canBuildCUIDot(formId: string, days: number) {
```
mathio, line 117: Its just a dot, can work for both CUI or form. It is not coupled to rendered interface:
```suggestion
function canBuildNotificationDot(formId: string, days: number) {
```

mathio (changes requested):

packages/demo-html/public/chat.html
antoniofull: I am fine with any naming

packages/embed/src/factories/create-popover/create-popover.ts
antoniofull: typo!

packages/embed/src/factories/create-popover/create-popover.ts
antoniofull: Just got it clarified by Max, I thought it was only for chat

packages/embed/src/factories/create-popover/create-popover.ts
antoniofull: I am mocking localStorage in the tests

packages/embed/src/factories/create-popover/create-popover.spec.ts
```diff
@@ -124,5 +134,25 @@ describe('#createSidetab', () => {
...
+      })
+
+      it('should hide notificationDot on form open', () => {
+        popover = createPopover('formId', { notificationDays: 2 })
+        fireEvent.click(screen.getByTestId('typeform-popover-button'))
+        jest.runAllTimers()
+        expect(screen.queryByTestId('typeform-popover-unread-dot')).toBeNull()
+      })
+
+      it('should store #notificationDot data in localStorage', () => {
```
mathio, line 151: 👍

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -97,6 +97,47 @@ const buildTooltip = (message: string, hide: () => void) => {
...
+
+type LocalStorageCUIData = {
+  formId: string
+  expireDate: number
+}
+const today = new Date().getTime()
+
+const cuiLocaslStorageData = localStorage.getItem(LOCAL_STORAGE_NOTIFICATION_DATA)
+
+const getLocalStorageData = () => cuiLocaslStorageData && JSON.parse(cuiLocaslStorageData)
```
mathio, line 110: Typo in var name. I would move it into `getLocalStorageData` method if possible:

```suggestion
const getLocalStorageData = () => {
  const data = localStorage.getItem(LOCAL_STORAGE_NOTIFICATION_DATA)
  return data && JSON.parse(data)
}
```

packages/embed/src/factories/create-popover/create-popover.ts
```diff
@@ -97,6 +97,47 @@ const buildTooltip = (message: string, hide: () => void) => {
...
+
+function saveToLocalStorage(data: LocalStorageCUIData) {
+  if (!data) return null
+  localStorage.setItem(LOCAL_STORAGE_NOTIFICATION_DATA, JSON.stringify(data))
```
mathio, line 114: For not data the method returns `null` but otherwise `undefined`. I would make it consistent:
```suggestion
  data && localStorage.setItem(LOCAL_STORAGE_NOTIFICATION_DATA, JSON.stringify(data))
```

packages/embed/src/factories/create-popover/create-popover.ts
mathio: I think this will also solve the issue with server-side rendering (failing build for nextjs demo)

mrisoli: getItem returns <T> | null so you can feed that to JSON.parse and not have to check here, otherwise this function returns false | Object

packages/embed/e2e/spec/functional/chat.spec.ts
```diff
@@ -0,0 +1,27 @@
+describe(`Chat Popover`, () => {
+  before(() => {
+    cy.visit('/chat')
```
mathio, line 3: I see there are some failing tests because this returns 404. I think it should be `/chat.html`.

mathio (approved):

mrisoli (approved):

scarciofolomarco (approved):

PR #252, opened 2021-06-01
fix(DIST-923): Fix notification days option expiry date

mathio:
- use correct `<` operator to display the notification dot
- save settings per form
- also fixes issue with overflowing text in tooltip
  - after recent changes it was requested that we should always wrap the text even if it has no spaces, but that causes this bug:
    Screenshot 2021-06-01 at 12 02 36
  - since text without spaces is less likely to happen, it can be cut off:
    Screenshot 2021-06-01 at 12 03 20


prilutskiy (approved):
🎉

packages/embed/src/factories/create-popover/notification-days.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    it('should return true when expiry date is not set in local storage and set it in local storage', () => {
+      const data = {
+        otherFormId: {
+          expiryDate: 10,
+        },
+      }
+      localStorageMockhelper(data)
+
+      expect(canBuildNotificationDot('myFormId', 2)).toBeTruthy()
+      expect(setItemMock.mock.calls[0][0]).toBe('tfNotificationData')
```
scarciofolomarco, line 25:
```suggestion
      expect(setItemMock).toHaveBeenCalledWith('tfNotificationData')
```

packages/embed/src/factories/create-popover/notification-days.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      expect(setItemMock.mock.calls[0][1]).toMatch(/{"otherFormId":{"expiryDate":10},"myFormId":{"expiryDate":[0-9]+}}/)
```
scarciofolomarco, line 26: not really sure about this one
```suggestion
      expect(setItemMock).toHaveBeenCalledWith(expect.stringMatching(/{"otherFormId:{"expiryDate":10},"myFormId":{"expiryDate":[0-9]+}/)
```

scarciofolomarco (approved):

packages/embed/src/factories/create-popover/notification-days.spec.ts
mathio: It is called with 2 args and 2nd one needs to be checked via regex.

packages/embed/src/factories/create-popover/notification-days.spec.ts
mathio: Is the other approach better? Jest docs suggest my approach: https://jestjs.io/docs/mock-functions#using-a-mock-function

PR #257, opened 2021-06-07
feat(WEB-1974): Disabling scroll for popup and slider

italoborges:
[Jira Ticket - WEB-1974](https://jira.typeform.tf/browse/WEB-1974)

Disabling the scrollbar preventing users to scroll when the popup or slider is open.
After closing it the scroll is back to its initial state.

### DIFF
--- a/packages/embed/src/factories/create-popup/create-popup.ts
+++ b/packages/embed/src/factories/create-popup/create-popup.ts
@@ -86,6 +86,7 @@ export const createPopup = (formId: string, userOptions: PopupOptions = {}): Pop
   const open = () => {
     if (!isOpen(popup)) {
       container.append(popup)
+      document.body.style.overflow = 'hidden'
       setTimeout(() => {
         popup.style.opacity = '1'
       })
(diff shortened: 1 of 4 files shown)


antoniofull (approved):
I am not sure if this has any accessibility implications. Other than that LGTM

packages/embed/src/factories/create-popup/create-popup.ts
```diff
@@ -96,6 +97,7 @@ export const createPopup = (formId: string, userOptions: PopupOptions = {}): Pop
     if (isOpen(popup)) {
       popup.style.opacity = '0'
       wrapper.style.opacity = '0'
+      document.body.style.overflow = 'initial'
```
mathio, line 100: What is the scrolling of the website was intentionally disabled (`hidden`) at the beginning? Will this preserve the behavior or will it enable scrolling regardless of its setting before opening the popup? 

Maybe we should preserve the value when opening the popup and re-use it here?

mathio (approved):

PR #277, opened 2021-06-29
fix(#274): Better way to handle empty querystring variables

### LINKED ISSUE #274: Are PR to master branch/classic lib still accepted?
HI, I'm seeing some unexpected behaviour while working with transferableUrlParameters (data-transaferable-url-parameters) when the querystring param is not present.

The [master code](https://github.com/Typeform/embed/blob/master/src/core/utils/url-parameters-transfer.js#L17-L21) that handles search params has this code below:

```js
  transferableUrlParameters.forEach((transferableParam) => {
    if (!(transferableParam in queryStrings)) {
      queryStringsWithTransferedParams[transferableParam] = urlSearchParams[transferableParam]
    }
  })
```

So the method is forcing all params from `data-transaferable-url-parameters` to be set in `queryStringsWithTransferedParams` as `undefined` and later, in method `appendParamsToUrl` this resultant object will be transformed into a querystring keeping the `undefined` as `"undefined"` due to the usage of `encodeURIComponent`.

```js
const appendParamsToUrl = (url, params) => {
  const queryParameters = []
  const { query, origin, pathname, hash } = UrlParse(url, true) // new URL(url) in browser env
  const path = pathname.replace(/\/$/, '') // remove trailing slash
  const parameters = Object.assign({}, query, params)

  Object.keys(parameters).forEach((key) => {
    queryParameters.push(`${encodeURIComponent(key)}=${encodeURIComponent(parameters[key])}`)
  })

  return `${origin}${path}?${queryParameters.join('&')}${hash}`
};
console.log(appendParamsToUrl('http://localhost/', { contains: 'value', missing: undefined }));
// http://localhost?contains=value&missing=undefined
```

This can be fixed working in a change for any (or both) methods listed above (`transferUrlParametersToQueryStrings` and `appendParamsToUrl`), but, as I said, I'm not sure if you are maintaining this old/classic version 🤔

rafaelverger:
### `transferUrlParametersToQueryStrings`

Resolves #274

**Context:**
```js
window.location = 'http://localhost/?empty-param=&filled=value' 
const urlParameters = ['empty-param', 'filled', 'non-existent']
```

**Before:**
```js
console.log(transferUrlParametersToQueryStrings(urlParameters, {}))
// { 'empty-param': '', 'filled': 'value', 'non-existent': undefined }
```

**After:**
```js
console.log(transferUrlParametersToQueryStrings(urlParameters, {}))
// { 'empty-param': '', 'filled': 'value' }
```

-------------------

### `appendParamsToUrl`

**Before:**
```js
console.log(appendParamsToUrl('http://localhost/?foo=bar#some=thing', { 'undef': undefined }))
// http://localhost/?foo=bar&undef=undefined#some=thing

console.log(updateQueryStringParameter('undef', undefined, 'http://localhost/?foo=bar#some=thing'))
// 'http://localhost?foo=bar&undef=undefined#some=thing'
```

**After:**
```js
console.log(appendParamsToUrl('http://localhost/?foo=bar#some=thing', { 'undef': undefined }))
// http://localhost?foo=bar&undef=#some=thing

console.log(updateQueryStringParameter('undef', undefined, 'http://localhost/?foo=bar#some=thing'))
// 'http://localhost?foo=bar&undef=#some=thing'
```

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -29,7 +29,9 @@ export const appendParamsToUrl = (url, params) => {
   const parameters = Object.assign({}, query, params)
 
   Object.keys(parameters).forEach((key) => {
-    queryParameters.push(`${encodeURIComponent(key)}=${encodeURIComponent(parameters[key])}`)
+    queryParameters.push(
+      `${encodeURIComponent(key)}=${parameters[key] == null ? '' : encodeURIComponent(parameters[key])}`
+    )
   })
 
   return `${origin}${path}?${queryParameters.join('&')}${hash}`
--- a/src/core/utils/url-parameters-transfer.js
+++ b/src/core/utils/url-parameters-transfer.js
@@ -15,7 +15,7 @@ export const transferUrlParametersToQueryStrings = (transferableUrlParameters, q
   const urlSearchParams = urlSearchToParams(urlSearchString)
   const queryStringsWithTransferedParams = { ...queryStrings }
   transferableUrlParameters.forEach((transferableParam) => {
-    if (!(transferableParam in queryStrings)) {
+    if (!(transferableParam in queryStrings) && transferableParam in urlSearchParams) {
       queryStringsWithTransferedParams[transferableParam] = urlSearchParams[transferableParam]
     }
   })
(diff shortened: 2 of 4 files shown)


rafaelverger:
This should fix #274

rafaelverger:
@mathio I believe I was able to cover the described scenarios in #274 with the this PR :)

mathio (approved):
Looking good! Thank you for the PR. When the rest of the team reviews this we can merge it.

Feel free to ignore the `secrets-scan` check below, I think it fails because you are not part of our company.

antoniofull (approved):
LGTM. Thanks for the pr 🙇

PR #279, opened 2021-07-01
feat(DIST-704): Add shareGoogleAnalyticsInstance option

antoniofull:
Adds Share Analytics Instance to Embed
- [Ticket DIST-704](https://jira.typeform.tf/browse/DIST-704)
- Reference To initial implementation [PI-88](https://jira.typeform.tf/browse/PI-88)
- Original Context [PI-63](https://jira.typeform.tf/browse/PI-63)
- Initial [Pull Request](https://github.com/Typeform/embed/pull/108)


packages/embed/src/utils/create-iframe/create-iframe.ts
```diff
@@ -18,6 +19,13 @@ export const createIframe = (formId: string, type: EmbedType, options: CreateIfr
   window.addEventListener('message', getFormQuestionChangedHandler(embedId, options))
   window.addEventListener('message', getFormSubmitHandler(embedId, options))
 
+  if (options.shareGoogleAnalyticsInstance) {
+    window.addEventListener('message', (ev) => {
+      if (ev.data.type === 'form-ready') {
+        setupGoogleAnalyticsInstanceSharingFeature(iframe, embedId)
+      }
+    })
+  }
```
antoniofull, line 28: I am not sure this is the best approach, but previously we would call it when the form is ready

packages/embed/src/utils/get-share-analytics-instance.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    gaObject((tracker: { get: (clientId: string) => string }) => {
+      const gaClientId = tracker.get('clientId')
+      sendGaIdMessage(gaClientId)
+    })
+  } catch (e) {
+    // eslint-disable-next-line
+    console.error(`Whoops! You enabled the shareGoogleAnalyticsInstance feature but the google analytics object has not been found.
+      Make sure to include Google Analytics Javascript code before the Typeform Embed Javascript code in your page. Error: ${e}`)
+  }
```
antoniofull, line 27: Two issues here:
1. Typing: I tried adding the @types/google.analytics but could not solve it
2. How and what to test here?

packages/embed/README.md
```diff
@@ -71,31 +71,32 @@ Or from admin panel URL:
...
+| notificationDays             | number      | display red notification dot, hide for given number of days since popover is open (popover only)                           | `undefined`                                                   |
+| onReady                      | function    | fires when the form is loaded                                                                                              | `undefined`                                                   |
+| onSubmit                     | function    | fires when user submits the form                                                                                           | `undefined`                                                   |
+| onQuestionChanged            | function    | fires when user navigates between form questions                                                                           | `undefined`                                                   |
+| shareGoogleAnalyticsInstance | function    | Allows to share the Google instance of the page with the embedded typeform                                                 | `false`                                                       |
```
mathio: Please update the description to match:
```
| shareGoogleAnalyticsInstance | function    | shares Google Analytics instance of the host page with embedded typeform                                                 | `false`                                                       |
```

PS: I am intentionally not suggesting this as Github "suggestion" here. Please update it in your IDE and run prettier on the file :)

packages/embed/src/utils/create-iframe/create-iframe.ts
mathio: Please reuse `getFormReadyHandler` handle (refactor it to take callback as 2nd param instead of "options" object) for this. This will validate the embed id and listen to events from correct form. This is necessary when there are multiple embeds on single page.

packages/embed/src/utils/get-share-analytics-instance.ts
```diff
@@ -0,0 +1,28 @@
+const GA_TYPE_MESSAGE = 'ga-client-id'
+
+export const setupGoogleAnalyticsInstanceSharingFeature = (iframe: HTMLIFrameElement, embedId: string) => {
+  const sendGaIdMessage = (gaClientId: string) => {
```
mathio, line 4: I would move this function outside of `setupGoogleAnalyticsInstanceSharingFeature` and add `embedId` as param.

packages/embed/src/utils/get-share-analytics-instance.ts
```diff
@@ -0,0 +1,28 @@
+const GA_TYPE_MESSAGE = 'ga-client-id'
+
+export const setupGoogleAnalyticsInstanceSharingFeature = (iframe: HTMLIFrameElement, embedId: string) => {
```
mathio, line 3: Lets not have the function name longer than necessary:

```suggestion
export const setupGoogleAnalyticsInstance = (iframe: HTMLIFrameElement, embedId: string) => {
```

Also please rename the filename to `setup-google-analytics-instance.js`.

packages/embed/src/utils/get-share-analytics-instance.ts
mathio: When you extract `sendGaIdMessage` outside of the function you can unit test it sends the postMessage on mocked iframe.

You could also test if the main function sends the postMessage with data from mocked `gaObject`.

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -37,6 +37,7 @@ const mapOptionsToQueryParams = (type: EmbedType, embedId: string, options: UrlO
     'disable-tracking': options.disableTracking || options.enableSandbox ? 'true' : undefined,
     'disable-auto-focus': options.disableAutoFocus ? 'true' : undefined,
     '__dangerous-disable-submissions': options.enableSandbox ? 'true' : undefined,
+    'share-ga-instance': options.shareGoogleAnalyticsInstance ? 'true' : undefined,
```
prilutskiy, line 40: let's use the same name everywhere, so it's easier to remember.
Either `shareGaInstance` + `share-ga-instance` or `shareGoogleAnalyticsInstance` + `shareGaInstance`

packages/embed/src/utils/create-iframe/get-form-event-handler.ts
```diff
@@ -1,6 +1,12 @@
-import { ActionableOptions } from '../../base'
+import { ActionableOptions, UrlOptions } from '../../base'
+import { setupGoogleAnalyticsInstance } from '../'
 
-export const getFormReadyHandler = (embedId: string, options: ActionableOptions) => {
+export interface FormReadyOptionTypes extends ActionableOptions, UrlOptions {}
+
+export const getFormReadyHandler = (embedId: string, iframe: HTMLIFrameElement, options: FormReadyOptionTypes) => {
+  if (options.shareGoogleAnalyticsInstance) {
+    setupGoogleAnalyticsInstance(iframe, embedId)
```
prilutskiy, line 8: please do not call it here. It would be hard to guess that `getFormReadyHandler` performs this side effect of setting up the GA instance.

packages/embed/src/utils/setup-google-analytics-instance.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+      }))
+      window['GoogleAnalyticsObject'] = 'ga'
+      window['ga'] = GoogleAnalyticsObjectMock as any
+
+      jest.runAllTimers()
+      setupGoogleAnalyticsInstance(iframe, 'embed-id')
+      expect(GoogleAnalyticsObjectMock).toHaveBeenCalledTimes(1)
+    })
+  })
+})
```
prilutskiy, line 46: Thanks for adding tests ❤️

packages/embed/src/utils/build-iframe-src.ts
antoniofull: the old one was called like this, might it break?

packages/embed/src/utils/create-iframe/get-form-event-handler.ts
antoniofull: 🤔  I see but this needs to be called when the iframe is fully loaded, hence on `form-ready` (and the `setupGoogleAnalyticsInstance ` needs the iframe passed to it). Where would it be the best place to load it?

packages/embed/src/utils/build-iframe-src.ts
mathio: Old v0 code will not work with v1 lib. It is acceptable to break compatibility.

packages/embed/src/utils/create-iframe/get-form-event-handler.ts
mathio: I think I would add another listener to the `form-ready` event with proper name.

antoniofull: But where do we want to add this event? We agreed not in `create-iframe.ts`, we agreed not here ☝️

antoniofull: The previous version was using React so it was loaded in `componentDidMount` for each embed

packages/embed/src/utils/create-iframe/create-iframe.ts
```diff
@@ -14,10 +15,16 @@ export const createIframe = (formId: string, type: EmbedType, options: CreateIfr
...
   iframe.addEventListener('load', triggerIframeRedraw, { once: true })
 
-  window.addEventListener('message', getFormReadyHandler(embedId, options))
+  window.addEventListener('message', getFormReadyHandler(embedId, options.onReady))
   window.addEventListener('message', getFormQuestionChangedHandler(embedId, options))
   window.addEventListener('message', getFormSubmitHandler(embedId, options))
 
+  // Get GA instance
+  if (options.shareGoogleAnalyticsInstance) {
+    getFormReadyHandler(embedId, () => {
```
mathio, line 24: I think you need to add listener: `window.addEventListener('message', getFormReadyHandler(...))`

antoniofull: Yeah my bad

mathio: Can we add this to `callbacks.spec.ts` and mock it somehow? My suggestion would be to create a mock GA object in `callbacks.html` and see if correct `postMessage` is sent to iframe (if we can spy on that function, I am not sure).

antoniofull: you mean in the functional tests?

packages/embed/src/utils/build-iframe-src.ts
antoniofull: @prilutskiy this is going to break the renderer, the best solution here is to then change everything to be `share-ga-instance` not the other way around

prilutskiy (approved):

packages/embed/src/utils/create-iframe/create-iframe.ts
mathio: Yes.

mathio: I looked into it, I dont think we can listen to messages sent into the iframe. We can proceed without functional tests with just unit tests.

However please test this in production when its merged.

mathio (approved):

packages/embed/src/utils/create-iframe/create-iframe.ts
antoniofull: Yes I do have one form with the GA id, however the methods are tested with unit test

PR #285, opened 2021-07-05
feat(DIST-1002): Auto close

scarciofolomarco:


packages/embed/src/utils/create-iframe/get-form-event-handler.ts
```diff
@@ -1,18 +1,18 @@
-import { ActionableOptions } from '../../base'
+type cb = (ev?: any) => void
```
mathio, line 1: Please name this properly.

packages/embed/src/utils/set-auto-close.ts
```diff
@@ -0,0 +1,10 @@
+import { getFormSubmitHandler } from './create-iframe/get-form-event-handler'
+
+export const setAutoClose = (embedId: string, options: { autoClose?: number }, callback: () => void) => {
```
mathio, line 3: Lets not pass all options, just the necessary one. Also lets rename the callback since it is always `close` function:

```suggestion
export const setAutoClose = (embedId: string, autoClose: number, closeCallback: () => void) => {
```

packages/embed/src/utils/set-auto-close.ts
```diff
@@ -0,0 +1,10 @@
...
+  if (options.autoClose) {
+    window.addEventListener(
+      'message',
+      getFormSubmitHandler(embedId, () => setTimeout(callback, options.autoClose))
```
mathio, line 7: What happens when the `autoClose` has no value, eg `<div data-tf-popup="id" data-tf-auto-close></div>`? Will it close? What will be the default timeout in this case? I think it should be zero and close right away.

packages/embed/e2e/spec/functional/auto-close.spec.ts
```diff
@@ -0,0 +1,27 @@
+describe('Auto Close', () => {
+  it('should auto close the form after submit', () => {
+    const autoCloseTime = 2000
+
+    cy.visit('/autoclose.html')
```
mathio, line 5: Nice tests!

packages/embed/src/utils/create-iframe/get-form-event-handler.ts
scarciofolomarco: any suggestion since I cannot use `callback` or `function`

scarciofolomarco: using `callbackFn` by now

packages/embed/src/utils/load-options-from-attributes.ts
```diff
@@ -69,6 +69,8 @@ export const transformAttributeValue = (value: string | null, transformation: Tr
       return transformArray(value)
     case 'record':
       return transformRecord(value)
+    case 'integerOrBoolean':
```
mathio, line 72: Add unit test please.

mathio (approved):

packages/embed/src/utils/load-options-from-attributes.ts
scarciofolomarco: added

antoniofull (approved):

PR #299, opened 2021-07-27
fix(DIST-1089): Use important for width and height

scarciofolomarco:


mathio (approved):

prilutskiy:
`!important` 😭 😭 😭

prilutskiy:
Also, be extra careful with vw vh, Safari doesn't always like them.

scarciofolomarco:
> `!important` 😭 😭 😭

yeah, I'm not proud but can't think a way to ignore `styles="width: #px; height: #px;"`

packages/embed/src/css/includes/_fullsize.scss
```diff
@@ -0,0 +1,6 @@
+@mixin fullsize {
+  width: 100% !important;
+  height: 100% !important;
+  width: 100vw !important;
+  height: 100vh !important;
+}
```
prilutskiy, line 6: That is really good. We'll be able to switch to another solution once find it by updating just a single place in the code. Nice!

prilutskiy (approved):

PR #310, opened 2021-08-05
fix(DIST-1131): Allow comas in hidden fields

mathio:
Array and object attributes can have values with coma. It needs to be escaped with backward slash.

### DIFF
--- a/packages/embed/README.md
+++ b/packages/embed/README.md
@@ -106,8 +106,9 @@ Or from admin panel URL:
 - define options as data attributes with `data-tf-` prefix and dashes in name (eg. `disableAutoFocus` becomes `data-tf-disable-autofocus`)
 - set a boolean property to `true` by omitting attribute value, (eg. `<div ... data-tf-disable-footer></div>`
 - pass function name for callbacks, eg. `data-tf-on-ready="myReadyFunction"` if this function is available on global scope (eg. `window`)
-- to pass `string[]` use coma-separated string, eg. `transitiveSearchParams: ['foo', 'bar']` becomes `data-tf-transitive-search-params="foo,bar"`
-- to pass `object` pass coma-separated key=value pairs, eg. `hidden: { foo: "f", bar: "b" }` becomes `data-tf-hidden="foo=f,bar=b"`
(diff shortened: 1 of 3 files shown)


packages/embed/src/utils/load-options-from-attributes.spec.ts
```diff
@@ -87,6 +87,10 @@ describe('load-options-from-attributes', () => {
       expect(transformAttributeValue('a, b', 'array')).toEqual(['a', 'b'])
     })
 
+    it('should transform string with escaped comas as array', () => {
+      expect(transformAttributeValue('a, b, c\\,d', 'array')).toEqual(['a', 'b', 'c,d'])
```
mathio, line 91: There is `\\` in test, but thats because you need to escape `\` with another `\`. When you write this in HTML attribute you can use just single `\` (tested).

I wanted to write a test where I would create an HTML element and read the value from there to demonstrate it, but to write `\` into the attribute I would need to write it as `\\` anyway :D

prilutskiy: yeah, def looks weird :d

prilutskiy (approved):

packages/embed/README.md
```diff
@@ -108,6 +108,7 @@ Or from admin panel URL:
 - pass function name for callbacks, eg. `data-tf-on-ready="myReadyFunction"` if this function is available on global scope (eg. `window`)
 - to pass `string[]` use coma-separated string, eg. `transitiveSearchParams: ['foo', 'bar']` becomes `data-tf-transitive-search-params="foo,bar"`
 - to pass `object` pass coma-separated key=value pairs, eg. `hidden: { foo: "f", bar: "b" }` becomes `data-tf-hidden="foo=f,bar=b"`
+  - **Note:** since comas `,` are used as delimiter for each value you will need to escape them with backward slash, eg. `data-tf-hidden="foo=foo\,bar"`. In JavaScript you don't need to escape it.
```
antoniofull, line 111: I think it's `commas` not `comas`

mathio: Good catch :D :D

antoniofull (approved):

PR #317, opened 2021-08-16
fix(DIST-1161): Fix height on mobile

antoniofull:
Ticket [DIST-1161](https://jira.typeform.tf/browse/DIST-1161)

### Issue 
The height of the root container is set to `100%vh` when embedded in  in mobile, that height exceed the height of the device.
testable broken typeform on https://embedtestsite.000webhostapp.com/basic-typeform/

### Fix
Sets the max height of the root to be not greater than the window.innerHeight, only for mobile devices

### DIFF
--- a/packages/embed/src/css/popover.scss
+++ b/packages/embed/src/css/popover.scss
@@ -15,6 +15,8 @@
   right: 16px;
   z-index: 10001;
 
+
+
   &.open {
     max-width: 100%;
     min-height: 360px;
@@ -169,6 +171,7 @@
       bottom: 0;
       right: 0;
       @include fullsize;
+      max-height: -webkit-fill-available;
 
       .typeform-popover-close {
         display: block;
--- a/packages/embed/src/css/popover.scss
+++ b/packages/embed/src/css/popover.scss
@@ -15,8 +15,6 @@
   right: 16px;
   z-index: 10001;
 
-
-
   &.open {
     max-width: 100%;
     min-height: 360px;
--- a/packages/embed/src/css/popover.scss
+++ b/packages/embed/src/css/popover.scss
@@ -169,7 +169,6 @@
       bottom: 0;
       right: 0;
       @include fullsize;
-      max-height: -webkit-fill-available;
 
       .typeform-popover-close {
         display: block;
--- a/packages/embed/src/css/includes/_fullsize.scss
+++ b/packages/embed/src/css/includes/_fullsize.scss
@@ -3,4 +3,6 @@
   height: 100% !important;
   width: 100vw !important;
   height: 100vh !important;
+  max-height: -webkit-fill-available;
+  max-height: -moz-fill-available;
 }


maxprilutskiy:
How do we plan to prevent this issue from happening in the future?

antoniofull:
> How do we plan to prevent this issue from happening in the future?

The fix will work with any mobile browser. `window.innerHeight` will always give the **real right height** of the viewport taking in considerations the bottom bar that is added to mobile browsers, that includes future changes to the browsers

scarciofolomarco (approved):

packages/embed/src/factories/create-popover/create-popover.spec.ts
```diff
@@ -11,6 +11,17 @@ const mockedLocalStorage = {
...
+    writable: true,
+    value: jest.fn().mockImplementation((query) => ({
+      matches: false,
+      media: query,
+      onchange: null,
+      addEventListener: jest.fn(),
+      removeEventListener: jest.fn(),
+      dispatchEvent: jest.fn(),
+    })),
+  })
```
maxprilutskiy, line 24: Wow, looks smart, what does it do exactly?

packages/embed/src/utils/set-max-height.ts
```diff
@@ -0,0 +1,11 @@
+import { isMobile } from './mobile'
+
+export const setMaxHeight = (element: HTMLElement) => {
+  const mediaQuery = window.matchMedia('(max-width: 480px)')
+
+  if (isMobile() && mediaQuery.matches) {
```
maxprilutskiy, line 6: please test this condition in a unit test

maxprilutskiy (approved):

packages/embed/src/factories/create-popover/create-popover.spec.ts
antoniofull: we need to mock `matchMedia` because js-dom does not support it

packages/embed/src/utils/set-max-height.ts
antoniofull: yeah I wanted to, but it's hard to unit test this code since it uses height and window. The mobile util is already tested, let me see if I can add some test to this file

packages/embed/src/utils/set-max-height.ts
```diff
@@ -0,0 +1,11 @@
+import { isMobile } from './mobile'
+
+export const setMaxHeight = (element: HTMLElement) => {
+  const mediaQuery = window.matchMedia('(max-width: 480px)')
```
mathio, line 4: This seems a bit odd. Do we set height for screens under certain width? What if the device is rotated and has width > 480px?

antoniofull: I added this media query because I see that on mobile, only below 480 we use the full height, if the popup or popover are above 480px they are not shown as full page. But yeah maybe could be removed and we always apply the maxHeight always? (In share page I solved this using styled-comp)

antoniofull: Maybe we can get rid of the MQ at all and just leave the maxHeight? WDYT

antoniofull: @mathio  @maxprilutskiy we can use css for this one, there is a `fill-available` intrinsic value that we can use for the max-height, it works on webkit and with the prefix will work also on Firefox. We do have few tickets related to this issue so the CSS solution might be better than using JS

packages/embed/src/css/popover.scss
```diff
@@ -15,6 +15,8 @@
   right: 16px;
   z-index: 10001;
 
+
+
```
mrisoli, line 19: Maybe remove these extra lines?

mrisoli (approved):

scarciofolomarco (approved):
only happens on popover? maybe is worth adding it to the `fullsize` mixin

packages/embed/src/css/popover.scss
antoniofull: yes missed that thanks!

antoniofull:
> only happens on popover? maybe is worth adding it to the `fullsize` mixin

Actually I think it happens also with popups, let me check it (the initial issue was just for popover)

antoniofull:
> only happens on popover? maybe is worth adding it to the `fullsize` mixin

done 🙇

mrisoli (approved):

PR #362, opened 2021-08-19
feat(DIST-1121): Document referrer as source

scarciofolomarco:


antoniofull (approved):

mrisoli (approved):

mathio (changes requested):
We should not populate source with referrer of the host page. I think this PR should be done in renderer repo. In renderer the referrer header has the value of page that linked to typeform (or embedded it in iframe). The "source" value should not be the full referrer URL, renderer should extract domain from it.

scarciofolomarco:
> We should not populate source with referrer of the host page. I think this PR should be done in renderer repo. In renderer the referrer header has the value of page that linked to typeform (or embedded it in iframe). The "source" value should not be the full referrer URL, renderer should extract domain from it.

that's why I wanted you to check this, gotta contact you for more details

mathio:
@scarciofolomarco I think we can close this.

PR #371, opened 2021-08-30
fix(DIST-1220): Set correct z-index for fullscreen widget on mobile

mathio:


maxprilutskiy:
What would be the best way to prevent this from happening in future? Do you think we could have a test for this or not really?

maxprilutskiy (approved):

mathio:
I added an element with z-index 10000 to the demo. We can actually add it to visual tests too and see if the embed covers this element. Nice catch @maxprilutskiy 👍

PR #373, opened 2021-08-31
feat(DIST-1234): Customizable html for PopupButton

maxprilutskiy:
Hi,

I'm adding this feature to the API to allow customization of PopupButton wrapper HTML. Especially useful when, as a consumer, you already have your own elements structure, and design.

### DIFF
--- a/packages/demo-react/src/App.js
+++ b/packages/demo-react/src/App.js
@@ -22,7 +22,7 @@ function App() {
       </p>
       <div className="App-spacer" />
       <div className="App-container">
-        <PopupButton asHtml={null} id="moe6aa" style={{ padding: 12, fontSize: 16 }} size={80}>
+        <PopupButton id="moe6aa" style={{ padding: 12, fontSize: 16 }} size={80}>
           click to open in popup
         </PopupButton>
       </div>
--- a/packages/demo-react/src/App.js
+++ b/packages/demo-react/src/App.js
@@ -22,7 +22,7 @@ function App() {
       </p>
       <div className="App-spacer" />
       <div className="App-container">
-        <PopupButton id="moe6aa" style={{ padding: 12, fontSize: 16 }} size={80}>
+        <PopupButton asHtml={null} id="moe6aa" style={{ padding: 12, fontSize: 16 }} size={80}>
           click to open in popup
         </PopupButton>
       </div>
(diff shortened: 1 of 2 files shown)


packages/embed-react/src/components/make-button-component.tsx
```diff
@@ -26,7 +27,7 @@ const emptyEmbed: GenericEmbed = {
 }
 
 function makeButtonComponent<T>(createFn: CreateFn<T>, cssFilename: string) {
-  return ({ id, children, style = {}, className = '', ...props }: ButtonComponentProps<T>) => {
+  return ({ id, children, as = 'button', style = {}, className = '', ...props }: ButtonComponentProps<T>) => {
```
maxprilutskiy, line 30:
```suggestion
  return ({ id, children, as = 'span', style = {}, className = '', ...props }: ButtonComponentProps<T>) => {
```
Maybe we should even fallback to `<span />` or `<div />` ?

mathio: I think we need to fallback to `<button>` otherwise exiting implementations will be affected.

packages/demo-react/src/App.js
```diff
@@ -22,7 +22,7 @@ function App() {
       </p>
       <div className="App-spacer" />
       <div className="App-container">
-        <PopupButton id="moe6aa" style={{ padding: 12, fontSize: 16 }} size={80}>
+        <PopupButton asHtml={null} id="moe6aa" style={{ padding: 12, fontSize: 16 }} size={80}>
```
mathio, line 25: What is `asHtml` prop? If you want to make an example if `as` prop please add it as a separate example below this one. Keep this one without the `as` prop.

maxprilutskiy, line 25:
```suggestion
        <PopupButton id="moe6aa" style={{ padding: 12, fontSize: 16 }} size={80}>
```

mathio (approved):
You dont seem to have correct ticket number in your commit message. If this change is necessary for DIST-1193 ticket, please create a subtask for embed lib work.

packages/demo-react/src/App.js
maxprilutskiy: Updated, thanks!

PR #374, opened 2021-08-31
feat(DIST-1235): Add onClose callback for embed

tfontenla:
Ticket: https://jira.typeform.tf/browse/DIST-1235

Adding the option to pass an onClose callback to the embed. We need this in cha-ching to improve UX and allow tracking when the user closes the form.


mathio (approved):

mathio:
Thank you for this, the code looks good 🚀

scarciofolomarco (approved):

PR #377, opened 2021-09-01
feat(DIST-1237): Add information about the repository

rivajunior:


maxprilutskiy (approved):

mathio (approved):
This is a "chore" task, but in order to release to npm we need "feat" or "fix" in commit message to make sure it is released by semantic-release.

PR #413, opened 2021-11-05
feat(DIST-1336): Show http warning

maxprilutskiy:


mathio (approved):
My personal preference would be to make the tests less verbose, other than that it looks good!

```js
const scenarios = [
  { url: '...', expected: true },
  ...
]

scenarios.forEach(({url, expected}) => {
    it(`should return ${expected} for URL ${url}`, () => {
      setUrl(url)
      expect(getIsWarningNeeded()).toBe(expected)
    })
})
```

antoniofull (approved):

PR #420, opened 2021-11-25
fix(DIST-1431): Allow camera and mic for iframe

scarciofolomarco:


packages/embed/src/utils/create-iframe/create-iframe.ts
```diff
@@ -16,6 +16,7 @@ export const createIframe = (formId: string, type: EmbedType, options: CreateIfr
   iframe.src = src
   iframe.dataset.testid = 'iframe'
   iframe.style.border = '0px'
+  iframe.allow = 'microphone; camera'
```
maxprilutskiy, line 19: please add a unit test

maxprilutskiy: other than that - looks good to me :)

mrisoli (approved):

PR #422, opened 2021-11-29
fix(DIST-1432): Prevent text from being hidden by ellipsis in side tab

antoniofull:
Screenshot 2021-11-29 at 11 43 22

### DIFF
--- a/packages/embed/src/css/sidetab.scss
+++ b/packages/embed/src/css/sidetab.scss
@@ -37,9 +37,11 @@
   &-button {
     position: absolute;
     top: 50%;
-    transform: rotate(-90deg);
-    transform-origin: bottom left;
-    width: 200px;
+    left: -48px;
+    transform: rotate(-90deg) translateX(-50%);
+    transform-origin: left top;
+    min-width: 100px;
+    max-width: 540px;
     height: 48px;
     display: flex;
     align-items: center;


mathio:
Can we align this to the center of the form modal window? Maybe we also need min-width for when the text is short?

antoniofull:
> Can we align this to the center of the form modal window? Maybe we also need min-width for when the text is short?
> 
> 

Yep makes sense to me.
> Maybe we also need min-width for when the text is short?
The spacing does not seem correct here, looking at it

packages/embed/src/css/sidetab.scss
```diff
@@ -37,9 +37,11 @@
...
     position: absolute;
     top: 50%;
-    transform: rotate(-90deg);
-    transform-origin: bottom left;
-    width: 200px;
+    left: -48px;
+    transform: rotate(-90deg) translateX(-50%);
+    transform-origin: left top;
+    min-width: 100px;
+    max-width: 300px;
```
mathio, line 44: Since the sidetab modal window has height of 580px, I think we can go up to 540px for max-width here. Its not the nicest design, but I think we should enable our customers to have longer texts there. I dont see a reason we should limit this to 300px (and for non-tech users there is a character limit in shareinguito anyway). WDYT?

packages/embed/src/css/sidetab.scss
```diff
@@ -37,9 +37,11 @@
   &-button {
     position: absolute;
     top: 50%;
-    transform: rotate(-90deg);
-    transform-origin: bottom left;
-    width: 200px;
+    left: -48px;
+    transform: rotate(-90deg) translateX(-50%);
+    transform-origin: left top;
+    min-width: 100px;
```
mathio, line 43: This is nice!

mathio (approved):

packages/embed/src/css/sidetab.scss
antoniofull: To me it makes sense, I went with the 300px fixed width because I followed the 24 chars limit that is on the share page, but otherwise if not needed in this case, makes totally sense to give more width and space here 🚀

PR #423, opened 2021-11-29
feat(TYP-6029): Change CSS classnames prefixes

mathio:
This prevents conflicts with old embed SDK lib (when both libs are in one page).


scarciofolomarco (approved):

maxprilutskiy:
Love the idea! Why `v1` though, and not just `tf` ?

maxprilutskiy (approved):

mathio:
If we have a new version with breaking changes in future we can rename the class to v2 and forget about possible conflicts. Credit to @scarciofolomarco for the idea!

PR #430, opened 2021-12-07
Update license metadata

eflynch:
This project seems to be distributed with an MIT license now. However, tools like `license-checker` report that it is LGPL licensed because of the metadata does not match.


mathio (approved):
I missed this in the original PR when updating the license. Thank you very much for the update!

antoniofull (approved):

PR #436, opened 2021-12-16
fix(DIST-1237): ModalWindowOptions['autoClose'] can also be a boolean

eliseumds:
By the way, shouldn't these JSDoc declarations include `undefined`?

Also, isn't it safer to end those lines with a semi-colon? I've never seen types declared like that, without semi-colons. Even tsc outputs them in *.d.ts* files.

### DIFF
--- a/packages/embed/src/base/modal-window-options.ts
+++ b/packages/embed/src/base/modal-window-options.ts
@@ -8,11 +8,13 @@ export type ModalWindowOptions = {
   /**
    * Time (ms) until the embedded typeform will automatically close after a respondent clicks the Submit button.
    *
-   * @type {number}
+   * @type {number|boolean}
    */
-  autoClose?: number
+  autoClose?: number | boolean
   /**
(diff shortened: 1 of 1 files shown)


mathio:
@eliseumds thank you for noticing and opening a PR 👍

> By the way, shouldn't these JSDoc declarations include `undefined`?

Those properties are [optional](https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties).

> Also, isn't it safer to end those lines with a semi-colon? I've never seen types declared like that, without semi-colons. Even tsc outputs them in _.d.ts_ files.

Semicolon at the end of a line is in most cases (like this one) a matter of preference. At typeform we have decided to write code without trailing semicolons and its [a standard across all our repositories](https://github.com/Typeform/eslint-config-typeform/blob/master/prettier/index.js).

eliseumds:
@mathio wow that was fast, thanks!

PR #450, opened 2022-02-02
feat(DIST-1481): Accept text and svg markup as icon

scarciofolomarco:


mathio (commented):
Please update the readme and dev portal docs on `customIcon` (explain what values does it accept).

BTW would it also accept any element? Eg. `<div>🤘</div>`? What about react component (like `<Sidetab customIcon={<MyIcon />} />`)?

mathio (approved):

PR #466, opened 2022-02-16
feat(DIST-1482): Specify redirect target

mathio:
### Redirect Target

You can supply a target for typeform redirect ([on submit or via ending](https://www.typeform.com/help/a/redirect-on-completion-or-redirect-through-endings-360060589532/)) as `redirectTarget` or `data-tf-redirect-target`.

It works the same as [target for anchor HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target):

- `_parent` (default), opens in parent page
- `_self` opens in the same embedded iframe as your typeform
- `_top` opens in current browser tab (same as `_parent` unless there are multiple nested iframes)
- `_blank` opens in new tab, however it might be blocked by popup blockers. **Warning:** It is not working in Safari (both desktop and mobile) and triggers a popup warning in Chrome on Android.

### ⚠️ Warning
Target _blank is not working in Safari (both desktop and mobile) and triggers a popup warning in Chrome on Android. It works in Chrome and Firefox on desktop.


maxprilutskiy (approved):
Awesome!

robertoibarra:
Lovely!! Thank you!!! 

I'm using it now in my HTML like:
<button data-tf-redirect-target="_self" data-tf-slider=".....

The second Typeform opens up in the side slider where the original form popped up :)

mimartinez:
Hi,
But if I have more than two Typeform it doesn't work. When it goes from the second to the third it already opens in the parent.

Issue #458, opened 2022-02-10
React embed injects inline <style> block, violating strict (no `unsafe-inline`) CSP policies: nonce support


dominics:
## Issue

The React embed currently injects a `<style>` block into the page.

For compliance and regulatory reasons, I want to use a `Content-Security-Policy` header that disallows such `unsafe-inline` style blocks, to give me better protection against XSS. This means I need some other way of validating the injected block is safe under CSP, otherwise I get a CSP violation (even after allow-listing various other Typeform CDN locations):

> `Refused to apply inline style because it violates the following Content Security Policy directive: "style-src-elem https://example.com https://cdn.example.com 'self' 'nonce-AAAA=='". Either the 'unsafe-inline' keyword, a hash ('sha256-ndjK5nlCVs//jbHQidaMOkHq5umwfQehMTu2NCP4pjM='), or a nonce ('nonce-...') is required to enable inline execution.`

## Possible solutions

There are two mechanisms under CSP that support inline blocks:
 - You can allow-list a static hash of the content of the block
   - This works but is fragile: it doesn't support dynamic content and if the embed gets updated, we have to keep the hash up to date or it'll be blocked again
 - You can use a nonce
 
 ### Nonces and `__webpack_nonce__`
 
Very quick summary of how the nonce works: inline style and script blocks need a `nonce` attribute set to a unique value generated by the server-side (or whatever generates the page). The value changes each request: code with access to this value is safely granted the ability to inject dynamic inline blocks.
 
This means that frontend components that inject inline blocks need to be able to obtain the current nonce value from the page at runtime. Things like Webpack and Styled Components that need to do just injection of inline blocks support a special global called `__webpack_nonce__`; if this global is set, they use it as the value of the nonce attribute attached to any injected `<style>` or `<script>` tags.

The Typeform embed could do the same thing: this would allow an optional nonce attribute to be attached to the inline style component, allowing it to pass strict content-security policies.

mathio:
I like this 👍  I am looking at your PR now.

PR #469, opened 2022-02-18
fix(DIST-1525): read the nonce from a bundler-safe location

dominics:
## Purpose

As I mentioned [in this comment](https://github.com/Typeform/embed/issues/458#issuecomment-1043760286) I think I made a mistake in how I accessed the `__webpack_nonce__` global from within the component, because by the time we bundle it with Webpack, this global reference is lost/minified.

## Approach

By referring to `global.__webpack_nonce__` rather than the bare `__webpack_nonce__`, we tell Webpack that this reference shouldn't change name during bundling.

### Emitted Code

Before this code, for the getNonce() function, after bundling we would emit:

```js
function getNonce() {
    return  true ? __webpack_require__.nc : 0;
}
exports.default = getNonce;
```

And after this PR it looks like:

```js
function getNonce() {
    return typeof __webpack_require__.g.__webpack_nonce__ !== 'undefined' ? __webpack_require__.g.__webpack_nonce__ : null;
}
exports.default = getNonce;
```

## Testing

This wasn't caught by unit tests last time, because it occurs after packaging. 

I failed to test enough last time, so this time I've bundled and loaded this change into my full application, and I can now see it working as intended (Chrome hides the value when it correctly matches the CSP, so this is working perfectly):
image

Lesson learned, sorry about that!

### DIFF
--- a/packages/embed-react/src/components/inline-style.spec.tsx
+++ b/packages/embed-react/src/components/inline-style.spec.tsx
@@ -3,14 +3,6 @@ import { render, screen } from '@testing-library/react'
 
 import { InlineStyle } from './inline-style'
 
-declare global {
-  namespace NodeJS {
-    interface Global {
-      __webpack_nonce__: string
-    }
-  }
-}
-
 describe('<InlineStyle />', () => {
   it('should render inline CSS', () => {
     render(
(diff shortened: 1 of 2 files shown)


mathio (approved):

mathio:
Thank you for a fix!

PR #483, opened 2022-03-09
fix(TYP-6447): Slider close button and border radius on mobile

mathio:


antoniofull (commented):
All those numbers in css are multiple of 4, wouldn't be more clear to have them in a calc function so that if we have to come back to it, we modify it once?

mathio:
@antoniofull can you show an example of what you mean please?

antoniofull:
> @antoniofull can you show an example of what you mean please?

It is probably not worth using a function, but maybe more DRY like using custom properties (or sass variable)
```CSS
--border-radius: 8px 0 0 8px;
--large: 24px;
--small: 4px;
```

antoniofull:
Anyway is more a NIT comment tbh

scarciofolomarco (approved):

mrisoli (approved):

PR #485, opened 2022-03-15
fix(DIST-1564): Remove event listeners on unmount

scarciofolomarco:


packages/embed/src/utils/create-custom-launch-options.ts
```diff
@@ -26,27 +36,36 @@ const openOnScroll = (scrollThreshold: number, open: () => void) => {
...
-      break
+      if (value) {
+        return openOnScroll(value, open)
+      }
+      return emptyHandler
     default:
-      // do not open automatically
-      break
+      return emptyHandler
+    // do not open automatically
```
mathio, line 69: NIT: comment should be before return.

mrisoli (approved):

PR #503, opened 2022-06-08
feat(DIST-1607): Use opaque background color for widget when opened as popup on mobile

mathio:


packages/embed/src/factories/create-widget/create-widget.ts
```diff
@@ -66,22 +75,31 @@ export const createWidget = (formId: string, options: WidgetOptions): Widget =>
...
     const originalHeight = container.style.height
     const openPopup = () => {
       container.classList.add('tf-v1-widget-fullscreen')
+      if (options.opacity !== undefined) {
+        container.style.backgroundColor = backgroundColor
+      }
       autoResize()
       window.addEventListener('resize', autoResize)
     }
+    const onTheme = (data: any) => {
```
antoniofull, line 90: you might want to use `unknown` instead of `any`

packages/embed/src/factories/create-widget/create-widget.ts
```diff
@@ -66,22 +75,31 @@ export const createWidget = (formId: string, options: WidgetOptions): Widget =>
   }
 
   if (widgetOptions.enableFullscreen) {
+    let backgroundColor = ''
     const { container } = options
     const autoResize = makeAutoResize(container)
     const originalHeight = container.style.height
     const openPopup = () => {
       container.classList.add('tf-v1-widget-fullscreen')
+      if (options.opacity !== undefined) {
```
maxprilutskiy, line 84: and what if `null`?

maxprilutskiy (approved):
Looks good! Left a small comment.

packages/embed/src/utils/change-color-opacity.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  it('should change decimal opacity to specified value', () => {
+    expect(changeColorOpacity('rgba(255, 125, 0, 0.5)', 125)).toBe('rgba(255, 125, 0, 125)')
+    expect(changeColorOpacity('rgba(255, 125, 0, .250)', 125)).toBe('rgba(255, 125, 0, 125)')
+  })
+
+  it('should not change opacity for color without opacity', () => {
+    expect(changeColorOpacity('rgb(255, 125, 0)', 25)).toBe('rgb(255, 125, 0)')
+    expect(changeColorOpacity('#ff0000', 25)).toBe('#ff0000')
+  })
+})
```
antoniofull, line 21: wouldn't be more clear in a test.each loop?

antoniofull (approved):
just a few comments

packages/embed/src/factories/create-widget/create-widget.ts
mathio: Then I get `TS2571: Object is of type 'unknown'.` error on next line. How would I solve that?

packages/embed/src/utils/change-color-opacity.spec.ts
mathio: What do you have in mind? Something like this?

```
import { changeColorOpacity } from './change-color-opacity'

describe('#changeColorOpacity', () => {
  const scenarios = [
    {
      title: 'should change opacity to 255 by default',
      inputColor: 'rgba(255, 125, 0, 0)',
      inputOpacity: undefined,
      output: 'rgba(255, 125, 0, 255)',
    },

    {
      title: 'should change opacity to specified value',
      inputColor: 'rgba(255, 125, 0, 50)',
      inputOpacity: 25,
      output: 'rgba(255, 125, 0, 25)',
    },

    {
      title: 'should change decimal opacity to specified value',
      inputColor: 'rgba(255, 125, 0, 0.5)',
      inputOpacity: 125,
      output: 'rgba(255, 125, 0, 125)',
    },

    {
      title: 'should change opacity to specified value',
      inputColor: 'rgba(255, 125, 0, .250)',
      inputOpacity: 125,
      output: 'rgba(255, 125, 0, 125)',
    },

    {
      title: 'should not change opacity for color without opacity',
      inputColor: 'rgb(255, 125, 0)',
      inputOpacity: 25,
      output: 'rgb(255, 125, 0)',
    },

    { title: 'should change opacity to specified value', inputColor: '#ff0000', inputOpacity: 25, output: '#ff0000' },
  ]

  scenarios.forEach(({ title, inputColor, inputOpacity, output }) => {
    it(title, () => {
      expect(changeColorOpacity(inputColor, inputOpacity)).toBe(output)
    })
  })
})
```

antoniofull: yep 👍  WDYT?

packages/embed/src/factories/create-widget/create-widget.ts
mathio: It can be `undefined` or integer, [see transformInteger method](https://github.com/Typeform/embed/blob/dist1607/packages/embed/src/utils/load-options-from-attributes.ts#L31-L34).

packages/embed/src/utils/change-color-opacity.spec.ts
mathio: I think it adds more abstraction to the test and makes it harder to read. If you write multiple `expect`s it is something every developer understands. If you build an array like this it is not obvious at first look what is going on.

I think this approach would be valid for a simpler scenarios, eg. [like this one](https://github.com/Typeform/embed/blob/dist1607/packages/embed/src/utils/is-defined.spec.ts#L15-L23).

packages/embed/src/factories/create-widget/create-widget.ts
mathio: I dont see any benefit of `unknown` over `any` - you dont know the type either way. I'd rather type it properly- there are multiple callback methods like this in the code, I will create a technical ticket for this.

packages/embed/src/factories/create-widget/create-widget.ts
maxprilutskiy: 💯

PR #507, opened 2022-06-23
feat(PR-32): clippy lib implementation

scarciofolomarco:


packages/demo-html/public/clippy-html.html
```diff
@@ -5,10 +5,15 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <title>Static HTML Demo</title>
     <link rel="stylesheet" href="./clippy/cursor-text.css" />
-    <script src="./clippy/cursor-text.js"></script>
+    <link rel="stylesheet" type="text/css" href="./lib/css/clippy.css" media="all" />
```
mathio, line 8: I think we need to load this in embed lib: https://github.com/Typeform/embed/blob/main/packages/embed/src/utils/include-css.ts#L3

Also, I dont see clippy files in the PR.

PR #510, opened 2022-07-05
chore(TYP-7505): Update Cypress version and run GH actions using latest node versions

mathio:


packages/embed/cypress.config.js
```diff
@@ -0,0 +1,24 @@
...
+  eyesIsDisabled: false,
+  eyesFailCypressOnDiff: true,
+  eyesDisableBrowserFetching: false,
+  eyesPort: 61632,
+  e2e: {
+    // We've imported your old cypress plugins here.
+    // You may want to clean this up later by importing these.
+    setupNodeEvents(on, config) {
+      return require('./cypress/plugins/index.js')(on, config)
+    },
```
antoniofull, line 20: Do we have to import the plugins?

packages/embed/cypress.config.js
```diff
@@ -0,0 +1,24 @@
...
+    specPattern: './e2e/spec/**/*.cy.{js,jsx,ts,tsx}',
```
antoniofull, line 21: Is this a new standard from the latest version of Cypress? Shall we change it in other repo's too?

mathio: This is the new standard coming from Cypress v10, it comes with a migration wizard that did all the changes related to Cypress in this PR.

scarciofolomarco (approved):

PR #511, opened 2022-07-19
fix(DIST-1643): Change default sidetab icon size

mathio:
Before:
Screenshot 2022-07-19 at 11 23 12

After:
Screenshot 2022-07-19 at 11 34 29

### DIFF
--- a/packages/embed/e2e/applitools-utils.ts
+++ b/packages/embed/e2e/applitools-utils.ts
@@ -1,6 +1,7 @@
 import { screenSizeDesktop, screenSizeMobile, Viewport } from './cypress-utils'
 
 const eyesCheck = async (name: string, { width, height }: Viewport) => {
+  cy.wait(2000)
   cy.eyesOpen({
     appName: 'Embed - Cypress',
     ignoreCaret: true,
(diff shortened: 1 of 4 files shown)


scarciofolomarco (approved):

packages/embed/e2e/applitools-utils.ts
```diff
@@ -1,6 +1,7 @@
 import { screenSizeDesktop, screenSizeMobile, Viewport } from './cypress-utils'
 
 const eyesCheck = async (name: string, { width, height }: Viewport) => {
+  cy.wait(2000)
```
antoniofull, line 4: wouldn't be better to wait for the element? I believe Cypress has a 5 seconds default time

antoniofull (approved):

packages/embed/e2e/applitools-utils.ts
mathio: I tried, but It looks like the wait is best solution.

antoniofull: 👍

Issue #513, opened 2022-08-03
Autofocus not working correctly with Widget


maxcountryman:
It seems the `embed-focus` event mentioned in #380 is no longer available. What this means is the iframe is not autofocused when using the `Widget` component. How can this be addressed with current versions of the embed package?

iparr:
Conversely, I'm using: `data-tf-disable-auto-focus` in an HTML embed and it steals focus.

How does one stop auto-focusing?

antoniofull:
HI @maxcountryman @iparr thanks for reporting this, we are investigating it and we will push a fix as soon as possible. I will let you know once we have fixed it. Thanks!

antoniofull:
HI, we have released a fix for that issue yesterday, could you let us know if now auto-focus is working fine? Thanks!

iparr:
That seems to have done the trick, cheers.

antoniofull:
Great! I am closing the issue, feel free to reopen it if you still have problems, thanks for reporting it.

maxcountryman:
@antoniofull what version tag is the fix in?

maxcountryman:
Using the latest package on NPM, this issue is not resolved for me. @antoniofull can you please reopen?

maxcountryman:
This is still an issue using the latest available on NPM. Please reopen the issue.

Can you point me to the commit that should have addressed this?

maxcountryman:
@antoniofull if you could reopen this, that would be great. I'm not able to.

antoniofull:
@maxcountryman the fix was released in another package, can you explain me please what is that you cannot achieve? Thanks

maxcountryman:
@antoniofull the issue is described in more detail [here](https://github.com/Typeform/embed/issues/513#issue-1327372676). Just to recap, this means that the keyboard controls don't work until the user manually clicks on the form.

Can you share which package you're referencing and the specific commit? Thanks!

mathio:
Hi @maxcountryman we are aware that there was indeed an auto focus issue on August 3rd. The issue was not in the embed library, however in the typeform itself. However the issue has been resolved since. 

Are you still experiencing any issues with auto focus? If yes, can you show as an example of embedded typeform with the issue? Thanks.

maxcountryman:
@mathio can you please reference the commit that you believe addressed the issue?

maxcountryman:
[...]

PR #521, opened 2022-09-13
feat(DIST-1865): Add option to enable auto focus

mathio:
New option `autoFocus` / `data-tf-auto-focus` to enable auto focus. 
Each embed also returns `focus()` method to dispatch the focus event manually.


packages/embed/src/factories/create-widget/create-widget.ts
```diff
@@ -64,6 +64,17 @@ export const createWidget = (formId: string, options: WidgetOptions): Widget =>
     )
   }
 
+  if (widgetOptions.autoFocus) {
+    window.addEventListener(
+      'message',
+      getFormReadyHandler(embedId, () => {
+        setTimeout(() => {
```
mathio, line 71: `setTimeout` is necessary because of the loading screen, see https://github.com/Typeform/embed/issues/513#issuecomment-1225593756

antoniofull (approved):

mrisoli (approved):

PR #523, opened 2022-10-03
fix(DIST-1614): Add support for node 18, break node 12 support

mathio:
BREAKING CHANGE: Node 12 is no longer supported by OpenJS Foundation. We will no longer support it in embed libs as well. Node 18 moves to Active LTS on 2022-10-25.
https://github.com/nodejs/release#release-schedule

This will be released as major version to make sure it is not automatically updated in old projects using node 12.This commit is not backwards compatible for node 12 apps as per [SemVer 2.0.0](https://semver.org/#spec-item-8):
> Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. It MAY also include minor and patch level changes.


docs/react.md
```diff
@@ -10,6 +10,11 @@ nav_order: 20
 
 ## Installation
 
+Requirements:
+
+- node >= 14 (for node v12 use [v1.21.0](https://www.npmjs.com/package/@typeform/embed-react/v/1.21.0))
```
mathio, line 15: Should we put the version for node 12 here? Or should we not encourage it? :)

packages/embed/src/css/includes/_spinner.scss
```diff
@@ -11,7 +13,7 @@ $spinner-sides: 1 !default;
   $diameter: $spinner-diameter,
   $sides: $spinner-sides
 ) {
-  $spinner-margin: $diameter / 2px * -1px;
+  $spinner-margin: math.div($diameter, 2px) * -1px;
```
antoniofull, line 16: nice!

antoniofull (approved):

packages/embed/src/css/includes/_spinner.scss
mathio: FYI it is a breaking change in latest sass. Unless you change this, it produces a warning:
https://sass-lang.com/documentation/breaking-changes/slash-div

PR #530, opened 2022-11-09
chore(DIST-1970): Bump node to v18 and up github actions

arturszott-tf:


.github/workflows/release.yml
```diff
@@ -48,18 +48,18 @@ jobs:
...
       - name: Get yarn cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         id: yarn-cache
         with:
           path: |
             **/node_modules
             ~/.cache
-          key: ${{ runner.os }}-node-12-yarn-${{ hashFiles('**/yarn.lock') }}-${{ hashFiles('.github/workflows/**.yml') }}
+          key: ${{ runner.os }}-node-16-yarn-${{ hashFiles('**/yarn.lock') }}-${{ hashFiles('.github/workflows/**.yml') }}
```
mathio, line 62: Can we do this?

```suggestion
          key: ${{ runner.os }}-node-${{ node -v }}-yarn-${{ hashFiles('**/yarn.lock') }}-${{ hashFiles('.github/workflows/**.yml') }}
```

mathio (approved):

.github/workflows/release.yml
arturszott-tf: This would print very specific v16.13.1, I guess in change of the node version, yarn.lock would be changed as well.

PR #535, opened 2022-11-22
fix(DIST-1988): Pass query param indicating there is embed callback for ending button click

mathio:


packages/embed/src/utils/create-iframe/create-iframe.ts
```diff
@@ -78,4 +78,4 @@ export const createIframe = (formId: string, type: EmbedType, options: CreateIfr
   return { iframe, embedId, refresh, focus }
 }
 
-type CreateIframeOptions = UrlOptions & ActionableOptions & IframeOptions
+export type CreateIframeOptions = UrlOptions & ActionableOptions & IframeOptions
```
mrisoli, line 81: This export was added here but the type itself is not being used anywhere else in the PR

mrisoli (approved):

PR #536, opened 2022-11-24
chore(DIST-1754): Add demo index page redirect

mrisoli:


mathio (commented):
I think this will not work, because [yarn release:aws:deploy](https://github.com/Typeform/embed/blob/8b9cfd1421d50e33340482c1f3d290b5bb56fc02/packages/embed/package.json#L52) runs this command:

```
DEBUG=jarvis yarn run jarvis deploy --path ./build-aws --version next
```

The `--version next` deploys everything into subdirectory so the `index.html` would end up in https://embed.typeform.com/next/.

However be extra careful when deploying to `/` path, since old embed lives there and I think we do not want to overwrite it yet. Maybe we could overwrite the `index.html` by putting it into `master` branch where previous embed lib version lives?

PR #537, opened 2022-11-24
fix(DIST-1754): Update index html

mrisoli:


typeform-ops-gha:
*[BOT]* Preview available: https://embed.typeform.com/0612df864673913d756314182822cc7092885f5c/embed.js

mathio:
I would suggest to comment out the [npm release job](https://github.com/Typeform/embed/blob/master/.github/workflows/release.yml#L54-L63) and instead just run the [yarn post-release](https://github.com/Typeform/embed/blob/master/package.json#L36) directly.

mathio (approved):

PR #542, opened 2022-12-01
feat(DIST-1914): Forward all transitive params

scarciofolomarco:


mathio (approved):
Looks good. Did you also test with HTML snippet like `<div data-tf-widget="HLjqXS5W " data-tf-transistive-search-params></div>`?

PR #545, opened 2022-12-19
feat(DIST-1914): Hidden field value priority over transitive params

scarciofolomarco:


mrisoli (approved):

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -102,6 +102,7 @@ export const buildIframeSrc = (params: BuildIframeSrcOptions): string => {
     Object.entries(options.hidden)
       .filter(([, paramValue]) => isDefined(paramValue))
       .forEach(([paramName, paramValue]) => {
+        url.searchParams.delete(paramName)
```
mathio, line 105: What if hidden field value is empty string? It will overwrite transitive search param, correct? Or will share page not include hidden fields with empty value?

mathio (approved):

packages/embed/src/utils/build-iframe-src.ts
scarciofolomarco: transitive params will take priority over empty hidden field values, pushing the changes..

packages/embed/src/utils/is-defined.ts
```diff
@@ -1,3 +1,7 @@
 export const isDefined = <T>(value: T | undefined | null): value is T => {
+  if (typeof value === 'string') {
```
mathio, line 2: I would suggest against this. The purpose of this method is to check if the value is defined, which empty string is.

scarciofolomarco: what would be your suggestion? create another method?

mathio: If it is used in multiple places then yes. Or just do `isDefined(value) && value !== ''`. Or maybe even `!!value` might work for `string | undefined`.

mathio (approved):
👍

PR #546, opened 2022-12-20
feat(DIST-2005): @typeform/embed-react add the ability to pass custom ref to PopupButton

vandercloak:
Let me know if this is already possible with the react package and I am blind. 

But I have a use-case where there I need more than just the button click to open the popover. 

I didn't want to go too deep and create a popover with state that can be passed in (no-button), and instead wanted to keep a refactor recommendation minimal by just allowing a ref override on the existing component.

First PR here, so please let me know if there is anything else that I can provide!

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -98,6 +98,27 @@ Pass options as props to the component.
 </PopupButton>
 ```
 
+### Passing a custom ref
+
+For some custom use cases it may be convenient to open the pop programmatically (without the button being clicked). 
+
+To do this, pass a ref to the PopupButton component and then use `ref.current.open()` to trigger the popover to open. 
+
+Example: 
+```javascript
+import { emptyEmbed } from "@typeform/embed-react";
+// ...
+const ref = useRef(emptyEmbed)
+const openPopup = () => ref.current.open()
+// ...
+<PopupButton
+  ref={ref}
+  ...
+>
+  click to open
+</PopupButton>
+```
+
 ## What's next?
 
 Learn more about [Vanilla Embed Library](/embed/vanilla). Since React Embed Library is just a React wrapper for Vanilla Embed Library, all concepts apply here as well.
--- a/docs/react.md
+++ b/docs/react.md
@@ -100,7 +100,7 @@ Pass options as props to the component.
 
 ### Passing a custom ref
 
-For some custom use cases it may be convenient to open the pop programmatically (without the button being clicked). 
+For some custom use cases it may be convenient to open the popup programmatically (without the button being clicked). 
 
 To do this, pass a ref to the PopupButton component and then use `ref.current.open()` to trigger the popup to open. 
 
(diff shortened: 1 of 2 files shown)


vandercloak:
@scarciofolomarco I noticed that you where the last one to modify this component, so I thought you may be a good person to message here. 

~~Is this something that is currently possible without this refactor? If not any ideas the best way to get eyes a review on this?~~ Just realized I can use `document?.getElementById("typeform-btn")?.click?.()`. 

Still feels like this would be useful to avoid a dom look up and to stick to the react paradigm. Curious your thoughts.

Thanks!!

mathio:
Hello @vandercloak , welcome and thank you for your PR.

Sorry I had to update your PR title to include our internal ticket number.

You are right you can do it either with DOM lookup which is not ideal when writing React or you could build a custom button. Since this can be solved by passing custom ref, I think your solution makes sense.

mathio:
In the PR you mention you want to implement this for `PopoverButton`, but you actually implemented it for `PopupButton` and `SliderButton`. Is it just a typo in the PR? 

Did you mean to also implement it also in `makeInitializerComponent`? https://github.com/Typeform/embed/blob/976eaa9d693dac9e6c2b1171d89024b49a62f78f/packages/embed-react/src/components/make-initializer-component.tsx#L19-L29

vandercloak:
> In the PR you mention you want to implement this for `PopoverButton`, but you actually implemented it for `PopupButton` and `SliderButton`. Is it just a typo in the PR?

Typo 🤦‍♂️

> Did you mean to also implement it also in `makeInitializerComponent`?

No. But I can if you think that would be useful!

I addressed the type in the title and the commit. Let me know if there is followup needed!

mathio:
Thanks! There is no followup needed, I will merge your PR as is.

If you feel like implementing this for `makeInitializerComponent` too, feel free to open another PR, we will appreciate it.

If not, no worries, I will put this into our tech backlog and we will pick it up some time next year 😃

mathio (approved):

docs/react.md
```diff
@@ -98,6 +98,27 @@ Pass options as props to the component.
 </PopupButton>
 ```
 
+### Passing a custom ref
+
+For some custom use cases it may be convenient to open the pop programmatically (without the button being clicked).
```
mathio, line 103:
```suggestion
For some custom use cases it may be convenient to open the popup programmatically (without the button being clicked). 
```

vandercloak:
Thanks [mathio](https://github.com/mathio)!

vandercloak:
@mathio I was looking into implementing this for `makeInitializerComponent` and the ts types don't make sense for the same ref work here. The `CreateFn` that is passed into makeInitializerComponent returns an artificial ref and not a true react ref. 

Happy to continue to support next year, but this will need a convo to scope out the behavior/functionality that we want to add.

PR #552, opened 2023-01-04
fix: Fix race condition with hubspot integration script

hburbano:
Add async/await to buildIframeSrc, cascade async/await function calls Add pooling on to hubspot cokkie check

Closes TYP-8021

[TYP-8021]: https://typeform.atlassian.net/browse/TYP-8021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


typeform-ops-gha:
*[BOT]* Preview available with hash `3ec14ec6ebbee1eb9ea2bcdd0d873c1bf9cb36e1` [here](https://embed.typeform.com/3ec14ec6ebbee1eb9ea2bcdd0d873c1bf9cb36e1/embed.js).

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -80,7 +81,7 @@ const getBaseUrl = (formString: string, chat: boolean = false): URL => {
   )
 }
 
-export const buildIframeSrc = (params: BuildIframeSrcOptions): string => {
+export const buildIframeSrc = async (params: BuildIframeSrcOptions): Promise<string> => {
```
mathio, line 84: I dont really like we are introducing a breaking change to the lib API just because of the hubspot functionality. It could break existing implementations that are using the lib from CDN. Is there another approach?

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -93,6 +94,9 @@ export const buildIframeSrc = (params: BuildIframeSrcOptions): string => {
     })
 
   if (options.hubspot) {
+    for (let maxRetries = 10, retries = 0; !getHubspotCookieValue() && retries < maxRetries; retries++) {
+      await sleep(250)
```
mathio, line 98: This waits for up to 2.5 seconds. Do we show any placeholder in the meantime?

mathio (changes requested):
Lets continue the conversation [in slack](https://typeform.slack.com/archives/C01K03A4EQ1/p1672905601247579).

PR #554, opened 2023-01-17
fix(DIST-2017): Race condition with hubspot integration script

mathio:
Closes https://github.com/Typeform/embed/pull/552

This fixes the race condition for when you embed via HTML API (eg `<div data-tf-widget="XX" data-tf-hubspot></div>`). For JavaScript API it is up to the 3rd party engineer to ensure they resolve any race conditions when create their embed.

It does not introduce any breaking changes.


packages/embed/src/initializers/initialize.ts
```diff
@@ -14,14 +14,18 @@ export const initialize = (
...
 
-  Array.from(embedTypeElements).forEach((element, index) => {
+  for (let index = 0; index < embedTypeElements.length; index += 1) {
+    const element = embedTypeElements.item(index)
     if (forceReload || element.dataset.tfLoaded !== 'true') {
       const formId = element.getAttribute(embedElementAttribute)
       if (!formId) {
         throw new Error(`Invalid ${embedElementAttribute}=${formId} for embed #${index}`)
       }
+      if (element.getAttribute('data-tf-hubspot') !== null) {
```
maxprilutskiy, line 24: how about `undefined`? Maybe worth changing to `!= null` to cover `undefined` as well?

typeform-ops-gha:
*[BOT]* Preview available with hash `57f5653b3bc2248e268045297efd394e3ed4cba7` [here](https://embed.typeform.com/57f5653b3bc2248e268045297efd394e3ed4cba7/embed.js).

packages/embed/src/utils/hubspot.ts
```diff
@@ -6,23 +8,19 @@ type HubspotFieldsType = {
...
- */
-
 export const getHubspotHiddenFields = (): HubspotFieldsType => ({
   hubspot_page_name: document.title,
   hubspot_page_url: window.location.href,
   hubspot_utk: getHubspotCookieValue(),
 })
+
+export const waitForHubspotCookie = async (): Promise<void> => {
+  for (let maxRetries = 10, retries = 0; !getHubspotCookieValue() && retries < maxRetries; retries++) {
```
maxprilutskiy, line 23: It looks like `getHubspotCookieValue()` will get called on every iteration, during the for condition evaluation, is that the idea? It's ok if it is, just thought I'd mention

maxprilutskiy (approved):

packages/embed/src/initializers/initialize.ts
hburbano:
```suggestion
      if (element.getAttribute('data-tf-hubspot')) {
```
An alternative here could be checking straightforward for a [Truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) value

mathio: When the attribute is not present the [return value is `null`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#non-existing_attributes). However if it is present, the value is an empty string and therefore [Falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) :D

Such is a life with JavaScript and browser APIs 😛

packages/embed/src/utils/hubspot.ts
mathio: Yes, thats the idea. It is a (hopefully quick) regex check. We get to break out from the "sleep cycle" as soon as the cookie is present.

mrisoli: Could this be done with [cookies.onChanged?](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged)

mathio: @mrisoli I see it is currently [not supported in Safari](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged#browser_compatibility) 😢

typeform-ops-gha:
*[BOT]* Preview available with hash `40dfab5c5524cc6efcbf82f702208d4c8bb9c65d` [here](https://embed.typeform.com/40dfab5c5524cc6efcbf82f702208d4c8bb9c65d/embed.js).

mrisoli (approved):

PR #556, opened 2023-01-20
feat(DIST-1929): Improve full screen embed experience

mathio:
Resize on screen resize. Also when browser navbars are displayed on mobile.

The idea is to be able to embed fullscreen with as simple code as this:

```html
<div data-tf-widget="form-id" data-tf-full-screen></div>
```

This will make fullscreen embedding easier when there already are any elements such as navigation bar that should be preserved. [Just set z-index of elements that should be displayed over typeform to >= 10001.](https://github.com/Typeform/embed/tree/main/packages/embed#positioning-and-overlapping)

# TODO

- [x] update README and docs


scarciofolomarco (approved):

typeform-ops-gha:
*[BOT]* Preview available with hash `9bc5a0696a2fcf892530a0f56fc0bf879cbc6564` [here](https://embed.typeform.com/9bc5a0696a2fcf892530a0f56fc0bf879cbc6564/embed.js).

PR #564, opened 2023-03-02
feat(PROS-217): Accept domain to run against different environment

zdenekkostal:
Accept an optional domain parameter to allow running against different environments.

Also fixes type issues in the tests by ignoring Cypress types to prevent the Assertion type leaking into Jest tests, screaming about matchers not being available.


typeform-ops-gha:
*[BOT]* Preview available with hash `481e4d61f825c3b4d9cffc6aff2fc12996b1056d` [here](https://embed.typeform.com/481e4d61f825c3b4d9cffc6aff2fc12996b1056d/embed.js).

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -71,20 +71,21 @@ const mapOptionsToQueryParams = (
...
 
-const getBaseUrl = (formString: string, chat: boolean = false): URL => {
+const getBaseUrl = (formString: string, hostname: string, chat: boolean = false): URL => {
   const prefix = chat ? 'c' : 'to'
+
   return new URL(
     formString.startsWith('http://') || formString.startsWith('https://')
       ? formString
-      : `${FORM_BASE_URL}/${prefix}/${formString}`
+      : `https://${hostname}/${prefix}/${formString}`
```
mathio, line 80: I think we should support hostnames with protocol as well, in case you want to use eg. `http://localhost`. WDYT?

mathio (commented):
This looks good 👍 Please let me know WDYT about the one comment below.

Please also update the table with all available options in both [README](https://github.com/Typeform/embed/tree/main/packages/embed#options) and [docs](https://github.com/Typeform/embed/blob/main/docs/configuration.md#available-options).

packages/embed/src/utils/build-iframe-src.ts
zdenekkostal: I was thinking that for the time being we could still continue with passing the complete URL via the form id (just for the testing purposes) and remove the part completely once we implement `yarn preview` with HTTPS in Jarvis to avoid having a code just for the testing. Alternatively, we could provide `baseUrl` and specify the whole part even with the protocol.

WDYT @mathio? cc @trapped @lussn @Coralie-TF

mathio: This works for me 🚀

trapped (approved):
Lovely, thanks @zdenekkostal!

typeform-ops-gha:
*[BOT]* Preview available with hash `bae3370ab90cf3f62410447d4e5f713dc2f8ad2e` [here](https://embed.typeform.com/bae3370ab90cf3f62410447d4e5f713dc2f8ad2e/embed.js).

packages/embed/src/utils/build-iframe-src.ts
```diff
@@ -71,20 +71,23 @@ const mapOptionsToQueryParams = (
...
-const getBaseUrl = (formString: string, chat: boolean = false): URL => {
+const getBaseUrl = (formString: string, domain: string, chat: boolean = false): URL => {
   const prefix = chat ? 'c' : 'to'
-  return new URL(
-    formString.startsWith('http://') || formString.startsWith('https://')
-      ? formString
-      : `${FORM_BASE_URL}/${prefix}/${formString}`
-  )
+  const domainWithProtocol =
+    domain.startsWith('http://') || domain.startsWith('https://') ? domain : `https://${domain}`
```
mathio, line 77: I like you avoided regex for the sake of readability 👍

typeform-ops-gha:
*[BOT]* Preview available with hash `bfcb97ec7784f4fd3a72084b5c93905cbe7951e7` [here](https://embed.typeform.com/bfcb97ec7784f4fd3a72084b5c93905cbe7951e7/embed.js).

PR #591, opened 2023-06-22
feat(REACH-458): Bump @typeform dependencies

gierlas:
### What

JIRA: [REACH-458](https://typeform.atlassian.net/browse/REACH-458)

- [x] `@typeform/eslint-config: ^6.0.0`

[REACH-458]: https://typeform.atlassian.net/browse/REACH-458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


typeform-ops-gha:
*[BOT]* Preview available with hash `f559e794ecd596e7d921db87258129e0bd4cea7b` [here](https://embed.typeform.com/f559e794ecd596e7d921db87258129e0bd4cea7b/embed.js).

packages/embed/src/utils/is-open.spec.ts
```diff
@@ -1,7 +1,10 @@
 import { isInPage, isOpen, isVisible } from './is-open'
 
 describe('is-open', () => {
-  const makeFakeElement = (parentNode: any, display?: string) =>
+  const emptyParentNode = {} as ParentNode
+  const undefinedParentNode = undefined as unknown as ParentNode
+
+  const makeFakeElement = (parentNode: ParentNode, display?: string) =>
```
mumpo, line 7: What if you type this as 

```ts
  const makeFakeElement = (parentNode: ParentNode | null, display?: string)

  makeFakeElement(null, 'inline-block')
```

That would be the same as this:

```ts
  const makeFakeElement = (parentNode: HTMLElement['parentNode'], display?: string)
```

And then replace `undefined` for `null`.

packages/embed-react/src/components/inline-style.spec.tsx
```diff
@@ -36,6 +36,7 @@ describe('<InlineStyle />', () => {
       </>
     )
 
+    // eslint-disable-next-line prefer-destructuring
     const styleElement = screen.getByTestId('style').getElementsByTagName('style')[0]
```
mathio, line 40: How about destructuring instead?
```suggestion
    const [ styleElement ] = screen.getByTestId('style').getElementsByTagName('style')
```

packages/embed-react/src/utils/nonce.ts
```diff
@@ -1,9 +1,6 @@
 declare global {
-  namespace NodeJS {
-    interface Global {
-      __webpack_nonce__: string
-    }
-  }
+  // eslint-disable-next-line no-var
+  var __webpack_nonce__: string
```
mathio, line 3: Can this be defined as `let` instead?
```suggestion
  let __webpack_nonce__: string
```

mathio (approved):

packages/embed-react/src/utils/nonce.ts
gierlas: I tried with let but then TS complains that it cannot find `__webpack_nonce__`. I guess the globals have to use `var`.

packages/embed-react/src/components/inline-style.spec.tsx
gierlas: Tried and got `Type 'HTMLCollectionOf<HTMLStyleElement>' is not an array type.` from Typescript 😞 Not sure why, because I don't see that complaint in the IDE 🤔

typeform-ops-gha:
*[BOT]* Preview available with hash `8aba62a8c6e22d5ace51bd7e96329a305949c94e` [here](https://embed.typeform.com/8aba62a8c6e22d5ace51bd7e96329a305949c94e/embed.js).

PR #612, opened 2023-08-30
fix(REACH-763): Remove height auto resize

mathio:
We no longer need to listen to "resize" events since Safari bug (https://bugs.webkit.org/show_bug.cgi?id=141832) was fixed earlier this year.


mathio:
For testing, make sure to use the JS API and load the CSS file separately to make sure the PR version of both CSS and JS files is loaded.

For example:
```html
<link rel="stylesheet" href="//embed.typeform.com/b4f222cf5b4096b422b5af832f4aa0bfee912bf4/css/widget.css"

<div id="form"></div>
<script src="//embed.typeform.com/b4f222cf5b4096b422b5af832f4aa0bfee912bf4/embed.js"></script>
<script>
  window.tf.createWidget("Cqrg7cgL", {
    container: document.querySelector("#form"),
    fullScreen: true  
  });
</script>
```

Thr44 (approved):

PR #632, opened 2024-01-02
fix(TU-7496): Require TS >=4.4.0

### LINKED ISSUE #631: Typescript error for partial-element-with-additional-attributes.d.ts
Screenshot 2023-12-27 093501

I believe the error stems from the syntax `[key: data${string}]: string;` being invalid because TypeScript does not support template literals as keys in index signatures. There seems to be an issue with the TypeScript definitions provided by this library. Is there a workaround for this?

mumpo:
Require Typescript version 4.4.0 in `@typeform/embed` by raising a deliberate TS error when using the package in lower versions.

This is necessary to use modern features such as Template Literals as object keys (see this issue: https://github.com/Typeform/embed/issues/631)

Resolves #631

### DIFF
--- a/packages/embed/package.json
+++ b/packages/embed/package.json
@@ -27,7 +27,9 @@
   "scripts": {
     "clean": "rm -rf ./build ./build-aws ./types",
     "build": "yarn webpack",
-    "postbuild": "tsc -p tsconfig.json --emitDeclarationOnly --declaration --declarationDir types --skipLibCheck",
+    "build:types": "tsc -p tsconfig.json --emitDeclarationOnly --declaration --declarationDir types --skipLibCheck",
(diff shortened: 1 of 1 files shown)


typeform-ops-gha:
*[BOT]* Preview available with hash `486095fba95966a58410f9a6ccb82363650f56dd` [here](https://embed.typeform.com/486095fba95966a58410f9a6ccb82363650f56dd/embed.js).

mathio (commented):
I installed the `@typeform/embed` in an empty project with `typescript@3`, however I did not get any warning about incorrect typescript version. However I did get a TS error:
```
node_modules/@typeform/embed/types/base/partial-element-with-additional-attributes.d.ts:3:11 - error TS1110: Type expected.

     [key: `data${string}`]: string;
```

Is this only for when building the library? I think that is not necessary, as in that case the correct version is installed as specified in `package.json` of the embed library.

mathio (approved):

PR #644, opened 2024-03-14
feat(TU-9756): support no-scrollbars flag

PabloMDiez:
Hi there! This piece of work intends to fix a glitch that can be observed with the following conditions:

- Windows OS (seem to always render very thick scrollbars)
- Chromium browsers
- `autoResize` on

When changing the parent window width, the typeform will recalculate sizes and word-wrapping, which can get the user in the following loop:

- Width changed
- Text elements in view cannot fit: some elements are moved to a new line
- Scrollbars are rendered immediately
- Width changes again (scrollbars are occupying space in view)
- If embed's height is smaller than the current block: resize
- Upon resize: scrollbars are no longer needed
- Scrollbars disappear
- Text can now fit with the extra width they now get: a newline is not needed
- Height changes: embed resizes

In some cases, this is reproducible in Chrome over Mac as well, but I got the worst glitch testing it in Windows 10 w/Chromium 122 & derivatives


typeform-ops-gha:
*[BOT]* Preview available with hash `b2acf2cd4e7258653246d21563d2d1df952eb12d` [here](https://embed.typeform.com/b2acf2cd4e7258653246d21563d2d1df952eb12d/embed.js).

mathio (approved):

PR #706, opened 2025-12-05
chore: add tmp new RX tests

joycevherck:
All the embed e2e and visual tests are currently excluded from the new renderer RX. However, the RX has been rolled out to 99% of users and we don't want to be completely in the dark. Because of that I'm suggesting we add just one duplicate test which tests a form that's been specifically added to the feature flag.

### DIFF
--- a/packages/demo-html/public/widget-newui-html.html
+++ b/packages/demo-html/public/widget-newui-html.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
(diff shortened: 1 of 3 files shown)


typeform-ops-gha:
*[BOT]* Preview available with hash `0b679f255235a2555247352ea607930993501290` [here](https://embed.typeform.com/0b679f255235a2555247352ea607930993501290/embed.js).

robespmun (approved):

PR #712, opened 2025-12-10
feat(TU-33149): Use an explicit --registry flag for jarvis install

tom-ridsdale:
The workflow was setting a global registry override for @typeform scoped packages:
`npm config set @typeform:registry https://npm.pkg.github.com/`

Idea is that this stayed in effect during the release, causing `semantic-release` to check authentication against GitHub Packages instead of npmjs.org, resulting in the error.

https://typeform.slack.com/archives/C01HXMZGU9E/p1765375968262809?thread_ts=1765366708.689729&cid=C01HXMZGU9E


tom-ridsdale:
Sonarcloud issues un-related.

PR #714, opened 2025-12-10
feat(TU-33149): Revert to previous jarvis install method but cleanup the registry override

tom-ridsdale:
Reverting to the previous method of installing `jarvis` given I know that works. The alternative, shorter approach is causing more problems. 

Difference to the original implementation is that we now clean-up the registry override which should hopefully solve the semantic-release issues (the original problem that the shorter approach was attempting to resolve).


typeform-ops-gha:
*[BOT]* Preview available with hash `c59bf27efec124f9cc12052eab8d90e50a03989d` [here](https://embed.typeform.com/c59bf27efec124f9cc12052eab8d90e50a03989d/embed.js).

PR #716, opened 2025-12-11
feat(TU-33149): Further cleanup

tom-ridsdale:
- Bumped `@semantic-release/exec` from `^6.0.3` to `^7.1.0`
- Ensure Get yarn cache commands are using node version 22 - This prevents using cached node_modules built for a different Node version.
- Noticed in the last release that it was analysing 800+ commits because `semantic-release-monorepo` was not loading - I think as a result of the invalid JSON in `.releaserc`.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -70,7 +70,7 @@ jobs:
           path: |
             **/node_modules
             ~/.cache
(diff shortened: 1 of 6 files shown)


typeform-ops-gha:
*[BOT]* Preview available with hash `bd49f4378bf25dd3692d0077a08177908569c631` [here](https://embed.typeform.com/bd49f4378bf25dd3692d0077a08177908569c631/embed.js).

PR #723, opened 2025-12-12
feat(TU-33149): Move to a new semantic release configuration file structure

tom-ridsdale:
Current problems
- NPM packages published with main dist-tag instead of latest
- GitHub releases marked as "Pre-release" instead of regular releases

After playing around with the dry-run functionality of `semantic-release` locally I found:
- In a monorepo, `semantic-release` runs from each package directory (e.g., packages/embed/), not from the repo root.
- cosmiconfig stops searching at the first `package.json` it finds, so it never reached the root `.releaserc`. This is not an issue in single-package repos where the config lives alongside the only package.json.
- In a previous PR, the `extends: ["semantic-release-monorepo"]` was moved from `.releaserc` to a `-e` command line flag. This change assumed the root config would still be found, but due to the monorepo structure, it wasn't.
- Therefore `semantic-release` used its default config, main wasn't the first branch in the list (maintenance patterns and master come first). This meant:
  - `Channel` defaulted to the branch name (`main`) instead of `null`
  - `main: false` was set (only the first branch in the list gets `main: true`)
  - This caused npm to use --tag main and GitHub to mark releases as pre-release.

Changes
- Rather than trying to maintain the old/legacy structure, move to a new structure.
- Created a shared config file at the repo root (`release.config.base.js`) with:
    - Valid JavaScript syntax (not JSON, so no quoting issues)
    - `extends: ['semantic-release-monorepo']` included
    - channel: false on the main branch
- Each package has a minimal `release.config.js` that imports the shared config, ensuring cosmiconfig finds it immediately when running from that package's directory.
- Removed the `-e` flag from package scripts since `extends...` is now in the base config file.


joycevherck (approved):

tom-ridsdale:
@joycevherck Would you mind approving again? & Are you comfortable with me bypassing the failure?

